PR #3, opened 2021-12-01
Create security.txt in .wellknown folder.

s-l-teichmann:
Create security.txt in the right folder.


s-l-teichmann:
Obsolete because of PR#5

PR #4, opened 2021-12-01
Improve handling of openpgp keyserver default

bernhardreiter:
* Use a default public keyserver from the new emerging keyservers
   network that sync each other, listed from https://spider.pgpkeys.eu/ .
 * Rename "PGP" to "OpenPGP" in the url parameter to refer to the open
   standard RFC 4880.


s-l-teichmann:
Branch static-metadata contains this change. Therefore closing here.

PR #21, opened 2021-12-21
Add visual representation for the success/fail status of requirements

Fadiabb:
* Add "ok" attribute to mark the status of the check
 * Adjust template to use css-entities for better status representation


cmd/csaf_checker/report.go
```diff
@@ -13,6 +13,7 @@ type Requirement struct {
 	Num         int      `json:"num"`
 	Description string   `json:"description"`
 	Messages    []string `json:"messages,omitempty"`
+	Ok          bool     `json:"ok"`
```
s-l-teichmann, line 16: @Fadiabb Spell it OK not Ok

PR #25, opened 2022-01-12
Configure nginx for client certificate authentication

Fadiabb:


docs/client-certificate-setup.md
```diff
@@ -0,0 +1,67 @@
...
+If the certificate Authority and user certificates are already present the steps of creating the certificate authority and client certificates can be skipped.
+The following is an example of creating them.
+
+```bash
+cd /etc/ssl
+```
+### Create the Certificate Autority (CA)
+
+Firstly, generate the CA:      
+```openssl genrsa -des3 -out ca.key 4096```
```
bernhardreiter, line 12: Even when it is an example, we should only use ciphers that are recommened for new systems in  BSI's TR-02102-1 (https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TG02102/BSI-TR-02102-1.html)    -des3 is not , -aes256 is

docs/client-certificate-setup.md
```diff
@@ -0,0 +1,67 @@
...
+```openssl genrsa -des3 -out ca.key 4096```          
+This asks to enter a passphrase.       
+Next, create the server-side certificate, that will be sent via the TLS server to the client.             
+```openssl req -new -x509 -days 365 -key ca.key -out ca.crt```       
+You will be asked to answer a few questions.
+
+### Create a client certificate
+
+Create the key like previously:              
+```openssl genresa -des3 -out userA.key 4906```
```
bernhardreiter, line 21: Typo genresa

bernhardreiter:
We consider using ECC (if browsers allow) and gnutls for the scripted creation of key material.

bernhardreiter:
Or even GnuPG for the creation of the certificates for something completely different. :) (But I'm not sure if this is documented, maybe doing a go application is even faster than doing this. ;) )

docs/client-certificate-setup.md
```diff
@@ -0,0 +1,67 @@
...
+
+```bash
+cd /etc/ssl
+```
+### Create the Certificate Autority (CA)
+
+Firstly, generate the CA:      
+```openssl genrsa -aes256 -out ca.key 4096```          
+This asks to enter a passphrase.       
+Next, create the server-side certificate, that will be sent via the TLS server to the client.
```
tschmidtb51, line 14: Is that the CA cert or the server cert? If the latter, in most cases that will be an existing server certificate or from Let's Encrypt. Please take that into consideration and clarify/differentiate which steps are needed when using an existing server cert.

docs/client-certificate-setup.md
bernhardreiter: (Infos now in other document)

bernhardreiter:
This branch now also implements the checking for tls certificates in the `csaf_provider` cgi binary.

What is missing here or elsewhere is to be more clear in the configuration about where the document root for nginx is and should be.

cmd/csaf_provider/controller.go
```diff
@@ -69,15 +70,24 @@ func (c *controller) bind(pim *pathInfoMux) {
...
-	}
 	return func(rw http.ResponseWriter, r *http.Request) {
-		hash := r.Header.Get("X-CSAF-PROVIDER-AUTH")
-		if !c.cfg.checkPassword(hash) {
+
+		verify := os.Getenv("SSL_CLIENT_VERIFY")
+		log.Printf("SSL_CLIENT_VERIFY: %s\n", verify)
+
+		switch {
+		case verify == "SUCCESS" && os.Getenv("SSL_CLIENT_I_DN") == c.cfg.Issuer:
```
bernhardreiter, line 79: Where is the documentation for this config options, especially the format it accepts.

bernhardreiter: @Fadiabb please add that documentation.

PR #30, opened 2022-01-18
Add Makefile for building go components

Fadiabb:
* Makefile with the following targets
	** build for (linux system)
     	** build_win  (windows system)
     	** build_tag (build from the last tag)
     	** clean  for (removing the binaries)
* Adjust README file

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 SHELL=/bin/bash
 BUILD = go build
-MAKDIR = mkdir -p binaries
+MKDIR = mkdir -p bin
 
 .PHONY: build build_win build_tag clean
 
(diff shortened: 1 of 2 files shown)


bernhardreiter:
To add some detail: The windows binaries target should be a cross build.

bernhardreiter:
@Fadiabb It should be possible to build and GNU/lInux and Windows binaries in one go, so they can be left in a subdirectory. 

And I think the tag build should ideally make use of the rules for the two binaries

s-l-teichmann:
Mhm ....

My expectation would me that a plain `make` builds the native target on the respective platform.
To trigger the cross build I would call it like `GOARCH=windows make` or alike

But as starting point I we should merge this in its current form.

Issue #33, opened 2022-01-24
CSAF uploader docs


tschmidtb51:
Please align the table with the `help` function. E.g.
- The `--insecure` flag seems to exist in the `csaf_uploader` but is not documented in the table.
- `-x` gives more info in the help than the table has

bernhardreiter:
Thinking about this: Do we need to duplicate the `--help` output in the README or elsewhere in the documentation?  

### cons
* If we did not, it would be easier to maintain.
### pros
* The documentation can be seen even if the tools is not build.
* It is possible to add additional infos there.

My current opinion is to remove the infos there and point towards running `--help` instead.
It is easy to run the tools and on tags we'll public binaries, so this should not be much of a huddle.
Besides the details only make sense if I can run the tools anyway.


And if there is more information to be added, we should generate it from the source code using godoc.

sthagen:
Personally, I prefer a README that contains a code block with the help command output over any manually created table. But then, I am not the table type.

PR #38, opened 2022-01-25
Improve uploader documentation

Fadiabb:
* #33

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,6 @@
 
 **WIP**: A proof of concept for a CSAF trusted provider, checker and aggregator.
 
-
 ## Setup
 
 - A recent version of **Go** (1.17+) should be installed. [Go installation](https://go.dev/doc/install)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,23 +2,24 @@
...
 - Clone the repository `git clone https://github.com/csaf-poc/csaf_distribution.git `
 
 - Build Go components
- ``` bash
- cd csaf_distribution
- go build -v ./cmd/...
+
+```bash
+cd csaf_distribution
+go build -v ./cmd/...
```
tschmidtb51, line 15: Maybe, we should explicitly create a `./bin` and put the executables in there? :thinking:

sthagen: I would keep it that way `./cmd/...` - it is common practice and bin is more where it ends up after installation (the best URL I could come up with using my current search-enging-foo is https://github.com/golang-standards/project-layout ...)

bernhardreiter:
@Fadiabb can you bring this pull-request up to current main, so it can be merged?

PR #40, opened 2022-01-27
Document provider config options

Fadiabb:


docs/provider-setup.md
```diff
@@ -85,9 +91,34 @@ domain = "http://192.168.56.102"
...
+   These affects the list items in the web interface.      
+   Default: `["csaf", "white", "amber", "green", "red"]`.
+ - upload_signature: Send signature with the request, an additional input-field in the web interface will be shown to let user enter an ascii armored signature. Default: `false`.
+ - openpgp_url: URL to OpenPGP key-server. Default: `https://openpgp.circl.lu`.
+ - domain: The domain name. Default: `http://$SERVER_NAME`.
+ - no_passphrase: Let user send password with the request, if set to true the input-field in the web interface will be disappeared. Default: `false`.
+ - no_validation: Validate the uploaded CSAF document against the JSON schema. Default: `false`.
+ - no_web_ui: Disable the web interface. Default: `false`.
+ - dynamic_provider_metadata: Take the publisher from the CSAF document. Default: `false`.
+ - publisher: Set the publisher. Default: `{"category": "vendor", "name": "ACME", "namespace": "https:/example.com"}`.
```
tschmidtb51, line 123: See comment about using ACME (ACME is a real company. Please use Example instead):
```suggestion
 - publisher: Set the publisher. Default: `{"category": "vendor", "name": "Example", "namespace": "https://example.com"}`.
```

Of course, that must be changed in the code as well

bernhardreiter:
Again I think, we may improve the point by generating the documentation, if this is easily possible.
See #33 for a similar question.

bernhardreiter:
As for implementing the documentation as part of the config code, e.g. in cmd/csaf_provider/config.go`: The toml library does not yet support handling of comments, see https://github.com/BurntSushi/toml/pull/328

docs/provider-setup.md
```diff
@@ -85,9 +91,34 @@ domain = "http://192.168.56.102"
...
+ - key: The private OpenPGP key.
+ - folder: Specify the root folder. Default: `/var/www/`.
+ - web: Specify the web folder. Default: `/var/www/html`.
+ - tlps: Set the allowed TLP comming with the upload request (one or more of "csaf", "white", "amber", "green", "red").             
+   The "csaf" selection lets the provider takes the value from the CSAF document.         
+   These affects the list items in the web interface.      
+   Default: `["csaf", "white", "amber", "green", "red"]`.
+ - upload_signature: Send signature with the request, an additional input-field in the web interface will be shown to let user enter an ascii armored signature. Default: `false`.
+ - openpgp_url: URL to OpenPGP key-server. Default: `https://openpgp.circl.lu`.
+ - domain: The domain name. Default: `http://$SERVER_NAME`.
```
bernhardreiter, line 118: @s-l-teichmann the `domain` config options is a bit unbalanced. It is called domain, but in the code, the value will be taken as base_url. And then there is `http:` which is okay for development, but not as default I guess. What about renaming it to serve_url and then default it to `https://$SERVER_NAME´ or so. Where does the name `SERVER_NAME` come from? My search engine does not tell me it is a standard environment variable. For development either `localhost` or `$hostname` would be alternatives. What do you think?

docs/provider-setup.md
bernhardreiter: Thanks for the question and the hint. We can use "Example" of course, though "ACME Corporation" seems to e a famous fictional company https://en.wikipedia.org/wiki/Acme_Corporation which is the ACME company that really exists (There was a car company, but spelled `Acme` ;) )

docs/provider-setup.md
bernhardreiter: @s-l-teichmann ping

s-l-teichmann: @bernhardreiter  We could use `https://` as a default, sure.
Concerning `domain` ... my bike shed opinion would more tend to something like `canonical_url_prefix`.

bernhardreiter: Okay, changed to `canonical_url_prefix`.

PR #51, opened 2022-02-03
Add some code documentation

Fadiabb:


cmd/csaf_provider/actions.go
```diff
@@ -129,6 +129,8 @@ func (c *controller) tlpParam(r *http.Request) (tlp, error) {
 	return "", fmt.Errorf("unsupported TLP type '%s'", t)
 }
 
+// creates calls the "ensureFolders" functions to create the directories and files.
```
sthagen, line 132: Nit: `creates` &rightarrow; `create`?

Fadiabb: Thanks for the hint.

PR #55, opened 2022-02-14
Add docs to run a development ca with gnutls

bernhardreiter:


bernhardreiter:
Took some inspiration from https://gist.github.com/epcim/832cec2482a255e3f392 though I reconstructed it. But I left a hint for people that find that recipe https://gist.github.com/epcim/832cec2482a255e3f392?permalink_comment_id=4064726#gistcomment-4064726 .

docs/install-server-certificate.md
```diff
@@ -14,8 +14,10 @@ There are three ways to get a TLS certificate for your HTTPS server:
  [Let's encrypt](https://letsencrypt.org/) without a fee.
  See their instruction, e.g.
  [certbot for nignx on Ubuntu](https://certbot.eff.org/instructions?ws=nginx&os=ubuntufocal).
- 3. Run your own little CA. Which has the major drawback that someone
- will have to import the root certificate in the webbrowsers manually.
+ 3. [Run your own little CA](development-ca.md).
+ Which has the major drawback that someone
+ will have to import the root certificate in the webbrowsers manually or
+ override warning on each connect.
  Suitable for development purposes.
```
tschmidtb51, line 21: I would like to add an explicit warning like "This should not be used in a production environment."

bernhardreiter: Addressed now. (But note that for client certificates it is more likely that a simple CA is used, even in production.)

PR #71, opened 2022-02-28
Use TLS client certificate for authentication (Checker)

Fadiabb:
* Add "client-cert" and "client-key" flag options to allow the checker to use TLS client certificate for authentication.

### DIFF
--- a/cmd/csaf_checker/processor.go
+++ b/cmd/csaf_checker/processor.go
@@ -219,15 +219,20 @@ func (p *processor) httpClient() *http.Client {
 	p.client = &http.Client{
 		CheckRedirect: p.checkRedirect,
 	}
-
+	var tlsConfig tls.Config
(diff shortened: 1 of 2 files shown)


cmd/csaf_checker/processor.go
```diff
@@ -775,7 +780,7 @@ func extractProviderURL(r io.Reader) (string, error) {
 		if strings.HasPrefix(line, csaf) {
 			line = strings.TrimSpace(line[len(csaf):])
 			if !strings.HasPrefix(line, "https://") {
-				return "", errors.New("CSAF: found in security.txt, but does not start with https://")
+				return "", errors.New("CASF: found in security.txt, but does not start with https://")
```
sebix, line 783: intentional change?

Fadiabb: No, thanks for the hint

PR #79, opened 2022-03-03
Improve Texts "PGP key" -> "public OpenPGP key"

bernhardreiter:


bernhardreiter:
@s-l-teichmann can you review and merge (should be uncontroversial)

Issue #66, opened 2022-02-25
Add --version option


bernhardreiter:
Would be nice to use `-v` and `--version` to see which version of the binary I have.

### technical considerations
We need one way to build this for released versions (if they are tagged and this is git we could take it from the tag)
and for development versions.

sebix:
Why `-v`? That is often used for verbose. `-V` would be another common short form of `--version`

bernhardreiter:
Good point, I guess `-V` is also fine. However https://www.gnu.org/prep/standards/standards.html#g_t_002d_002dversion does not have a standard short form for `--version`, maybe the long form is enough, as it is a rarely used options anyway.

bernhardreiter:
Implemented with #89

PR #89, opened 2022-04-07
Dev version flag

Fadiabb:
Add `-v` `--version` flag to display the version of "csaf_checker", "csaf_uploader" and "csaf_provider" components. 
If it is a release version, the tag name of the release will be displayed.
if it is not, the short version of the commit hash will be appended to "-dev" and the date of the commit to display.


cmd/csaf_checker/main.go
```diff
@@ -29,6 +31,7 @@ type options struct {
 	Insecure   bool    `long:"insecure" description:"Do not check TLS certificates from provider"`
 	ClientCert *string `long:"client-cert" description:"TLS client certificate file (PEM encoded data)" value-name:"CERT-FILE"`
 	ClientKey  *string `long:"client-key" description:"TLS client private key file (PEM encoded data)" value-name:"KEY-FILE"`
+	Version    bool    `long:"version" short:"v" description:"Display version of the binary"`
```
bernhardreiter, line 34: According to https://github.com/csaf-poc/csaf_distribution/issues/66#issuecomment-1055111262 we do not want to have a short version for this option.

cmd/csaf_uploader/main.go
```diff
@@ -48,7 +49,8 @@ type options struct {
 
 	Insecure bool `long:"insecure" description:"Do not check TLS certificates from provider"`
 
-	Config *string `short:"c" long:"config" description:"Path to config ini file" value-name:"INI-FILE" no-ini:"true"`
+	Config  *string `short:"c" long:"config" description:"Path to config ini file" value-name:"INI-FILE" no-ini:"true"`
```
bernhardreiter, line 52: One ` ` too much?

util/file.go
```diff
@@ -140,3 +142,44 @@ func mkUniq(prefix string, create func(string) error) (string, error) {
...
 	return "", err
 }
+
+// GetVersion returns the version of the binary file. It distinguishes between two cases:
+// a) if the current commit is the same commit of the last tag it returns the tag's name.
+// b) if the current commit differs from the commit of the last tag it returns
+// the short commit hash of the actual commit appended to the "dev" and date of commit.
+func GetVersion() (string, error) {
+	var err error
+	versionCmd := exec.Command("git", "describe", "--tag", "--abbrev=0")
```
bernhardreiter, line 152: This makes the binary depending on being able to call `git`, which is not desirable. The version number should be determined at build time and then "burned" into the binary.

cmd/csaf_checker/main.go
```diff
@@ -132,10 +135,19 @@ func main() {
...
 	errCheck(err)
 
-	if len(domains) == 0 {
+	if len(domains) == 0 && !opts.Version {
 		log.Println("No domains given.")
 		return
 	}
+	if opts.Version {
+		version, err := util.GetVersion()
+		if err != nil {
```
bernhardreiter, line 144: Couldn't error handling being part of util.GetVersion()? then the returned string could be printed in any case. This would reduce code duplication. (And there should be no case where getVersion fails, see the other remark.)

bernhardreiter:
> Why not always use git describe?

We need to change the approach anyway, something like
https://gitlab.com/sascha.l.teichmann/the-exile/-/blob/master/Makefile#L16 the -X method
like described here https://gitlab.com/sascha.l.teichmann/the-exile/-/blob/master/Makefile#L16

bernhardreiter:
@tschmidtb51 
> Why not always use git describe?

Took a look, it is a useful command, the only drawback being that it does not produce https://semver.org compatible version numbers. So we probably have to dissect its output and increase the PATCH level number, if we use it.

tschmidtb51:
> @tschmidtb51
> 
> > Why not always use git describe?
> 
> Took a look, it is a useful command, the only drawback being that it does not produce https://semver.org compatible version numbers. So we probably have to dissect its output and increase the PATCH level number, if we use it.

Sorry - I might be missing something: Do we need to have semver? There is no way an automated system (at least at the moment) can reliable detect the version changes meant be semver... So I'm happy to use `git describe` for anything in development and `semver` for tagged releases...

util/file.go
```diff
@@ -140,3 +150,19 @@ func mkUniq(prefix string, create func(string) error) (string, error) {
 
 	return "", err
 }
+
+// GetVersion returns the version of the binary file. It distinguishes between two cases:
+// a) if the current commit is the same commit of the last tag it returns the tag's name.
+// b) if the current commit differs from the commit of the last tag it returns
+// the short commit hash of the actual commit appended to the "dev" and date of commit.
+func GetVersion() (string, error) {
```
bernhardreiter, line 158: Should not compute the version number during runtime, this logic should be in the Makefile during the build and linking process. (And also increase the PATCH version part for dev Versions.)

cmd/csaf_checker/main.go
```diff
@@ -132,10 +135,14 @@ func main() {
 	domains, err := flags.Parse(opts)
 	errCheck(err)
 
-	if len(domains) == 0 {
+	if len(domains) == 0 && !opts.Version {
```
sthagen, line 138: From the sidelines: To me that smells a bit. Why not change the flow by first doing the version request dance and then one can spare relating surprisingly the domains and the request for version data dimensions.

cmd/csaf_uploader/main.go
bernhardreiter: (no it's fine because column aligned with later lines)

PR #90, opened 2022-04-07
Correct double line

tschmidtb51:
There was a weird wording. I guess it comes from resolving a merge conflict...


bernhardreiter:
@tschmidtb51 thanks for the correction! (We have not yet finalized all the documentation, so there will be more little things like this and we plan to fix them.)

Issue #97, opened 2022-04-08
Use CSD02 JSON schema


tschmidtb51:
We should use the CSD02's [JSON schema](https://github.com/oasis-tcs/csaf/blob/master/csaf_2.0/json_schema/csaf_json_schema.json).

PR #98, opened 2022-04-08
CSD02 Schema update

tschmidtb51:
- resolves csaf-poc/csaf_distribution#97
- add strict schemas from OASIS repo


bernhardreiter:
@s-l-teichmann fine with you?

PR #102, opened 2022-04-12
Pack README.md into the artifacts (Github action)

Fadiabb:


Fadiabb:
As these action generates several files for each binary, and there is no option to pack the binaries in one file, an issue is opened by: [https://github.com/wangyoucao577/go-release-action/issues/82](https://github.com/wangyoucao577/go-release-action/issues/82)

bernhardreiter:
One simple solution would be, that if the action can "make", to put the building of the the two distribution targets (one archive for GNU/Linux, one for windows) in the make file.

bernhardreiter:
We won't merge, but use a different route via Makefile.

PR #110, opened 2022-04-21
Update the security.txt

### LINKED ISSUE #35: Update security.txt
It would be nice, if the security.txt is updated (adding the CSAF field) if it already exists.

s-l-teichmann:
If there is already a security.txt file we add the CSAF line in front of it 
in case it is not already in.

Resolves #35

### DIFF
--- a/cmd/csaf_provider/create.go
+++ b/cmd/csaf_provider/create.go
@@ -40,7 +40,7 @@ func ensureFolders(c *config) error {
 		return err
 	}
 
-	return createSecurity(c, wellknown)
+	return setupSecurity(c, wellknown)
 }
 
 // createWellknown creates ".well-known" directory if not exist and returns nil.
@@ -86,9 +86,12 @@ func createFeedFolders(c *config, wellknown string) error {
 	return nil
 }
 
-// createSecurity creates the "security.txt" file if does not exist
-// and writes the CSAF field inside the file.
-func createSecurity(c *config, wellknown string) error {
+// setupSecurity creates the "security.txt" file if does not exist
+// and writes the CSAF field inside the file. If the file exists
+// it checks ig the CSAF entry with the provider-metadata.json
+// path is already in. If its not it is added in front of all lines.
+// Otherwise the file is left untouched.
+func setupSecurity(c *config, wellknown string) error {
 	security := filepath.Join(wellknown, "security.txt")
 
 	path := fmt.Sprintf(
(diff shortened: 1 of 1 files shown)


cmd/csaf_provider/create.go
```diff
@@ -87,20 +90,85 @@ func createFeedFolders(c *config, wellknown string) error {
 // and writes the CSAF field inside the file.
```
tschmidtb51, line 90: Please also correct the comment to reflect the new functionality.

cmd/csaf_provider/create.go
```diff
@@ -87,20 +90,85 @@ func createFeedFolders(c *config, wellknown string) error {
...
 func createSecurity(c *config, wellknown string) error {
```
tschmidtb51, line 91: Maybe the function should be renamed, e.g. in `setupSecurity` :thinking:

cmd/csaf_provider/create.go
```diff
@@ -87,20 +90,85 @@ func createFeedFolders(c *config, wellknown string) error {
...
+		return err
+	}
+
+	// we are already in the file.
+	if found {
+		return nil
+	}
+
+	// Insert our CSAF line at the beginning
+	// to get higher priority over possible existing CSAF lines.
```
tschmidtb51, line 142: I like that! :smile:

tschmidtb51 (commented):
I didn't test the it but the code looks good.

cmd/csaf_provider/create.go
s-l-teichmann: [x] done

cmd/csaf_provider/create.go
s-l-teichmann: [x] done

cmd/csaf_provider/create.go
```diff
@@ -87,20 +90,85 @@ func createFeedFolders(c *config, wellknown string) error {
...
+	csafLine := fmt.Sprintf("CSAF: %s", path)
+	lines = append([]string{csafLine}, lines...)
```
bernhardreiter, line 147: The examples in https://datatracker.ietf.org/doc/html/draft-foudil-securitytxt-12 suggest that an additional linefeed could make the file more readable. In addition we could consider adding a comment like:

```
# proposed CSAF-2.0 field. Inserted by csaf_provider/create on 2022-04-22
```

bernhardreiter:
Did a test, positive, so far:

* on create an existing security.txt file is updated if we are not in there.
* on an upload, there is no check

cmd/csaf_provider/create.go
s-l-teichmann: Could be done in separate PR.

bernhardreiter: Added the newline for now.

PR #114, opened 2022-04-27
Dev nginx dns

Fadiabb:
Add nginx configuration to serve `provider-metadata.json` according to the DNS 
`csaf.data.security.domain.tld` record.


bernhardreiter:
@Fadiabb the server name in the script does not work-out for tests, it must be a variable so it can be set to something that we could resolve like `csaf.data.security.localhost` which we also put into `/etc/hosts` for the integration tests.

bernhardreiter:
If this is the configuration of a virtual host that we want to keep, we could write it into a new file under `/etc/nginx/sites-available
` and link it to `sites-enabled`, this way it is more seperate.

PR #116, opened 2022-04-28
Moved code from checker to library.

s-l-teichmann:


s-l-teichmann:
To enable re-sync of main and aggregator branches.

Issue #83, opened 2022-03-25
Make provider configurable for being listed or mirrored


bernhardreiter:
The provider should be configurable for indicating that it wants to be
 * listed
 * mirrored

PR #118, opened 2022-04-29
For issue #83

s-l-teichmann:
The branch provider-listed-mirrorred2 implements this.
In the config there is a new optional table provider_metadata with
bool fields list_on_CSAF_aggregators and mirror_on_CSAF_aggregators
which are both are optional. If not set the defaults of the standard are used (true).

The publisher field of the config is moved into the provider_metadata table
as it is on the same level as list and mirror options in the generated provider_metadata.json
file.

### DIFF
--- a/docs/provider-setup.md
+++ b/docs/provider-setup.md
@@ -136,6 +136,9 @@ Provider has many config options described as following:
  - no_validation: Validate the uploaded CSAF document against the JSON schema. Default: `false`.
  - no_web_ui: Disable the web interface. Default: `false`.
  - dynamic_provider_metadata: Take the publisher from the CSAF document. Default: `false`.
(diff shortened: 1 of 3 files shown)


docs/provider-setup.md
```diff
@@ -136,6 +136,9 @@ Provider has many config options described as following:
  - no_validation: Validate the uploaded CSAF document against the JSON schema. Default: `false`.
  - no_web_ui: Disable the web interface. Default: `false`.
  - dynamic_provider_metadata: Take the publisher from the CSAF document. Default: `false`.
- - publisher: Set the publisher. Default: `{"category"= "vendor", "name"= "Example", "namespace"= "https://example.com"}`.
+ - provider_metadata: Configure the provider metadata.
+ - provider_metadata.list_on_CSAF_aggregators: List on aggregators
```
bernhardreiter, line 140: Default?

PR #124, opened 2022-05-10
Disable summary from fixed notes

bernhardreiter:
* Comment out compiling the summary and requiring document.notes to fix main provider.


bernhardreiter:
We won't merge, because we'll implement a better solution today.

Issue #94, opened 2022-04-07
Upload: 403 results in interesting error message


tschmidtb51:
When trying to upload without the `password`, it results correctly in 403. But the message seems a little bit weird:
```
Upload failed: 403 Forbidden
error: invalid character 'F' looking for beginning of value
```
I guess https://github.com/csaf-poc/csaf_distribution/blob/530a02742effff01e9183eabdfd467501153a25c/cmd/csaf_uploader/main.go#L289-L291 should exit the function if a 403 occurs...

bernhardreiter:
Yes the error handling can be improved in this case.

bernhardreiter:
We'll plan some code cleanup next.

s-l-teichmann:
This should be solved with PR  #247

JanHoefelmeyer:
It's indeed solved by https://github.com/csaf-poc/csaf_distribution/pull/247

PR #150, opened 2022-05-23
Handle "403" error (provider)

Fadiabb:
* #94

### DIFF
--- a/cmd/csaf_uploader/main.go
+++ b/cmd/csaf_uploader/main.go
@@ -296,6 +296,9 @@ func (p *processor) process(filename string) error {
 	if resp.StatusCode != http.StatusOK {
 		uploadErr = fmt.Errorf("upload failed: %s", resp.Status)
 		fmt.Printf("HTTPS %s\n", uploadErr)
+		if resp.StatusCode == http.StatusForbidden {
+			return err
+		}
 	}
 
 	var result struct {


cmd/csaf_uploader/main.go
```diff
@@ -296,6 +296,9 @@ func (p *processor) process(filename string) error {
 	if resp.StatusCode != http.StatusOK {
 		uploadErr = fmt.Errorf("upload failed: %s", resp.Status)
 		fmt.Printf("HTTPS %s\n", uploadErr)
+		if resp.StatusCode == http.StatusForbidden {
+			return err
```
bernhardreiter, line 300: @s-l-teichmann 
We can upload several files, should be try them all, even if we are getting errors on some or abort on the first error?

s-l-teichmann: @bernhardreiter Yes, maybe we should continue with the next one.

bernhardreiter: @s-l-teichmann right now, we do. Okay, then this remark is resolved.

cmd/csaf_uploader/main.go
```diff
@@ -296,6 +296,9 @@ func (p *processor) process(filename string) error {
 	if resp.StatusCode != http.StatusOK {
 		uploadErr = fmt.Errorf("upload failed: %s", resp.Status)
 		fmt.Printf("HTTPS %s\n", uploadErr)
+		if resp.StatusCode == http.StatusForbidden {
```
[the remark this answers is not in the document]
bernhardreiter: And we need to print out which file failed, if we did not get this back from the server.

bernhardreiter:
We plan to solve this with a different pull request.

PR #154, opened 2022-05-31
Provider metadata

s-l-teichmann:
Implement improved lookup for provider-metadata.json.


bernhardreiter:
@s-l-teichmann if the download fails, because of a certificate TLS problem, there is no message explaining this. Only that it failed.

PR #157, opened 2022-05-31
Remove redundant brackets.

s-l-teichmann:
Remove redundant brackets via `gofmt -r '(a) -> a' -w *.go`.


bernhardreiter:
Because && is evaluated before ||, the brackets are unnecessary (reference https://go.dev/ref/spec#Operator_precedence).

Issue #85, opened 2022-03-31
Place public OpenPGP key in the file system directory `openpgp` 


bernhardreiter:
As clarified from @tschmidtb51 and others in January, the public OpenPGP key used for signing the CSAF files shall be placed in a directory `openpgp/` next to the `provider-metadata.json` by the provider.

Accordingly this is where the `pgp_keys` url should point to.

bernhardreiter:
#### Rationale
Not using WKD or public pubkey servers is making the implementation more reliable and simpler for provider and client implementors. 

WKD would require an email address and public pubkeyservers often do not provide additional data to verify the pubkey.

PR #161, opened 2022-06-07
Openpgp folder

s-l-teichmann:
Create openpg folders with copies of the public keys. Solves issue #85.

### DIFF
--- a/cmd/csaf_aggregator/config.go
+++ b/cmd/csaf_aggregator/config.go
@@ -15,7 +15,6 @@ import (
 	"net/http"
 	"os"
 	"runtime"
-	"strings"
 	"sync"
 
 	"github.com/BurntSushi/toml"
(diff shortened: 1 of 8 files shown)


bernhardreiter:
It misses to update the docs for aggregator.

And integration test fails https://github.com/csaf-poc/csaf_distribution/runs/6793125642?check_suite_focus=true

with something like
```
    <strong>Error: <tt>open /var/www/html/.well-known/csaf/openpgp/39539fa0413d9c9c09ee16ae3640cadea01b4a2f.asc: no such file or directory.</tt></strong>
279
```
during the uploader --action create.

bernhardreiter:
@s-l-teichmann BTW: the liking of issue #85 and closing did not workout yet, maybe because you've used `issue` in `Solves isse #85`. Should possibly work when using `solves #85` ?!

bernhardreiter:
It seems a little bit more common to use upper case characters when using fingerprints.  Guess this would only be one change.

bernhardreiter:
The openpgp fingerprint example in the csaf 2.0 csd02 is also upper case, so we should change it.

s-l-teichmann:
For the peace of mind.

PR #162, opened 2022-06-08
Added types to checker messages.

s-l-teichmann:
Added types to checker messages.

### DIFF
--- a/cmd/csaf_checker/links.go
+++ b/cmd/csaf_checker/links.go
@@ -53,11 +53,11 @@ func (pgs pages) listed(path string, pro *processor) (bool, error) {
 	pro.badDirListings.use()
 
 	if err != nil {
-		pro.badDirListings.add(ErrorType, "Fetching %s failed: %v", base, err)
+		pro.badDirListings.error("Fetching %s failed: %v", base, err)
 		return false, errContinue
 	}
 	if res.StatusCode != http.StatusOK {
-		pro.badDirListings.add(ErrorType, "Fetching %s failed. Status code %d (%s)",
+		pro.badDirListings.error("Fetching %s failed. Status code %d (%s)",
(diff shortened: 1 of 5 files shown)


cmd/csaf_checker/links.go
```diff
@@ -53,11 +53,11 @@ func (pgs pages) listed(path string, pro *processor) (bool, error) {
 	pro.badDirListings.use()
 
 	if err != nil {
-		pro.badDirListings.add("Fetching %s failed: %v", base, err)
+		pro.badDirListings.add(ErrorType, "Fetching %s failed: %v", base, err)
```
bernhardreiter, line 56: Should .add() be replaced with .error() here (and four lines down) as well? (Just like in other files for consistency?)

s-l-teichmann: Yes, was an oversight. Corrected now, even if it was only a minor thing.

cmd/csaf_checker/tmpl/report.html
```diff
@@ -8,13 +8,13 @@
   <body>
     <h1>CSAF-Checker - Report</h1>
 {{- range .Domains }}
-    <h2>{{ .Name }}</h2>
+    <h2>{{ .Name }}{{ if .HasErrors }} (failed){{ end }}</h2>
 
     <dl>
 {{ range .Requirements }}
-    <dt><strong>Requirement {{ .Num }}: {{ .Description }}</strong></dt>
+    <dt><strong>Requirement {{ .Num }}: {{ .Description }}{{ if .HasErrors }} (failed){{ end }}</strong></dt>
```
bernhardreiter, line 15: "failed" does not fit perfectly, as the checker run was successful (in doing checks and finding stuff). So maybe `Problems found` or `found problems`?

Issue #75, opened 2022-03-02
Improve checking for new ROLIE possibilities


bernhardreiter:
There are forthcoming changes towards CSAF-2.0 cs2 and it would be nice if the checker could check for the improved ROLIE values (in the area of hashes and sigs).

#### maybe out of scope for 1.0

PR #181, opened 2022-06-14
Integrity check the ROLIE feed advisory URLs, hashes and signatures.

s-l-teichmann:
Solves #75

### DIFF
--- a/cmd/csaf_checker/processor.go
+++ b/cmd/csaf_checker/processor.go
@@ -329,8 +329,55 @@ func (p *processor) httpClient() util.Client {
 
 var yearFromURL = regexp.MustCompile(`.*/(\d{4})/[^/]+$`)
 
+// checkFile constructs the urls of a remote file.
+type checkFile interface {
+	url() string
+	sha256() string
+	sha512() string
+	sign() string
+}
+
+// stringFile is a simple implementation of checkFile.
+// The hash aund signature files are directly by exteding
+// the file name.
+type stringFile string
+
+func (sf stringFile) url() string    { return string(sf) }
+func (sf stringFile) sha256() string { return string(sf) + ".sha256" }
+func (sf stringFile) sha512() string { return string(sf) + ".sha512" }
+func (sf stringFile) sign() string   { return string(sf) + ".asc" }
+
+// hashFile is a more involed version of checkFile.
+// Here each component can be given explicitly.
+// If a component is not given it is constructed by
+// extending the first component.
+type hashFile [4]string
+
+func (hf hashFile) url() string { return hf[0] }
+
+func (hf hashFile) sha256() string {
+	if hf[1] == "" {
+		return hf[0] + ".sha256"
+	}
+	return hf[1]
+}
+
+func (hf hashFile) sha512() string {
+	if hf[2] == "" {
+		return hf[0] + ".sha512"
+	}
+	return hf[2]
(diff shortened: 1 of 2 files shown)


bernhardreiter:
One test, I've changed a value to
```json
"href": "https://localhost:8443/.well-known/csaf/white/2022/bsi-2022-0001.json.sha156"
```
and run the checker   "version": "0.9.2-15-ga179896" with `--verbose`, but got no message and the url above was also not tried.  Is this behaviour to be expected?

s-l-teichmann:
> 



> One test, I've changed a value to
> 
> ```json
> "href": "https://localhost:8443/.well-known/csaf/white/2022/bsi-2022-0001.json.sha156"
> ```
> 
> and run the checker "version": "0.9.2-15-ga179896" with `--verbose`, but got no message and the url above was also not tried. Is this behaviour to be expected?

Yes, as it ends with .sha156 ... only files ending with .sh256 and .sha512 are considered for check sums.

s-l-teichmann:
> ### second test
> 
> In another test I've renamed a signature file like this
> 
> ```shell
> mv /var/www/html/.well-known/csaf/white/2022/bsi-2022-0001.json.asc /var/www/html/.well-known/csaf/white/2022/bsi-2022-0001.json.asco
> ```
> 
> and changed the entry in ` /var/www/html/.well-known/csaf/white/csaf-feed-tlp-white.json` to match it.
> 
> The checker run tried the url, and reported no issues. This seems not perfectly right as https://docs.oasis-open.org/csaf/csaf/v2.0/csd02/csaf-v2.0-csd02.html#7119-requirement-19-signatures demand an "appropriate extension".

"rel" is "signature" in this case which is enough to identify it as a signature URL. The URL structure itself
is not check. I would prefer this to be checked in the schema, but we could add it in the checker.

> ### third test
> 
> Destroyed on hash, used a different signature file, checker run detected this:
> 
> ```
>     <dt><strong>Requirement 18: Integrity (failed)</strong></dt>
>     <dd>- ERROR: SHA256 hash of https://localhost:8443/.well-known/csaf/white/2022/bsi-2022-0001.json does not match https://localhost:8443/.well-known/csaf/white/2022/bsi-2022-0001.json.sha256.</dd>
> 
>     <dt><strong>Requirement 19: Signatures (failed)</strong></dt>
>     <dd>- ERROR: Signature of https://localhost:8443/.well-known/csaf/white/2022/bsi-2022-0001.json could not be verified.</dd>
> ```

Can you post what exactly is in your feed file, please.

s-l-teichmann:
> #### Re first test
> 
> > as it ends with .sha156 ... only files ending with .sh256 and .sha512 are considered for check sums.
> 
> However it is given as relative 'hash' and the rel `signature` is considered, beside the "unusual" filename in the **second test**. Wouldn't it be consistent to either consider "unusual" filenames in both tests or i non?
> 
> And if the choice is made that an "unusual" filename is detected, but not considered, shouldn't it be a warning?
> #### Re third test
> 
> > > Destroyed on hash, used a different signature file,  checker run detected this
> > > [..]
> > > Can you post what exactly is in your feed file, please.
> 
> Irrelevant as the test result was as expected: It did the test and found out that it did not match in case of the hash and the signature. Sorry that I was not clear enough about it.

It issues more warnings now. Please re-test.

Issue #126, opened 2022-05-11
Provide option to run mandatory tests


tschmidtb51:
With [secvisogram/csaf-validator-service](https://github.com/secvisogram/csaf-validator-service) a service exists that could run the business level test on CSAF documents and provide the information whether the document is valid. We should add a config option to send the CSAF document to such a service and parse the result. If `isValid` is true, the upload should proceed, otherwise an error should be thrown.
See [the documentation](https://github.com/secvisogram/csaf-validator-service/blob/main/documents/generated/asciidoc/index.adoc) for details on how to send the document and receive the response.

bernhardreiter:
Shall the service be run on the same server where the csaf_provider also runs?
Or is it an external service that the csaf_uploader should consult?

bernhardreiter:
@tschmidtb51 told me, the idea is to have an unauthenticated service that runs within someones own infrastructure and is called on the provider or aggregator side (optionally, when the URL is configured).

bernhardreiter:
Further clarifications:
 * It is fine that the nothing is mirrored, if in a mirror-run there is a validation error in at least one advisory.
 * An optimisation should be implemented, that compares the downloaded advisories to the previously downloaded ones and only uses the validation service for advisories that have not been validated in that instance before.

bernhardreiter:
(it is contracted and planned now)

PR #185, opened 2022-06-15
Add support for remote validation services.

s-l-teichmann:
Solves #126


bernhardreiter:
Would be nice to have a link to the remote service which can be used for this. And another one how to set the service up. (Ideally we add an integration test script for it.)

s-l-teichmann:
> Would be nice to have a link to the remote service which can be used for this. And another one how to set the service up. (Ideally we add an integration test script for it.)

I'm not aware of a public available service. I can only tell you what I've done to start such a service locally.
These steps came from the docs linked in #126

Issue #103, opened 2022-04-13
Provide useful aggregator functionality as module for a potential downloader implementation


bernhardreiter:
Some functions implemented in the `csaf_aggregator` are useful when implementing a downloader.
The should be provided that they can be used easily.

bernhardreiter:
(implementation of csaf_downloader is contracted and planned now)

PR #194, opened 2022-06-22
CSAF downloader

s-l-teichmann:
Add a tool to download advisories from providers.

Solves #103


bernhardreiter:
Found #197 while testing, but it is a general issue, so does not block merging.

bernhardreiter:
Usage does not say that domains can be given:
```bash
./bin-linux-amd64/csaf_downloader -h
Usage:
  csaf_downloader [OPTIONS]
[..]
```

bernhardreiter:
Giving a direct url does not work (yet):
```bash
~/csaf_distribution-git/bin-linux-amd64/csaf_downloader https://localhost:9443/.well-known/csaf-aggregator/local-dev-provider2/provider-metadata.json --insecure
2022/06/23 11:07:50 Looking for provider-metadata.json of 'https://localhost:9443/.well-known/csaf-aggregator/local-dev-provider2/provider-metadata.json': https://https://localhost:9443/.well-known/csaf-aggregator/local-dev-provider2/provider-metadata.json/.well-known/csaf/provider-metadata.json not found.
```

It seems there is something appended, where it shouldn't.

bernhardreiter:
We do not download and validate  the checksums and signatures yet, though we SHOULD according to https://docs.oasis-open.org/csaf/csaf/v2.0/csd02/csaf-v2.0-csd02.html#732-retrieving-csaf-documents ;)

Anyway, could download something. It builds and the integration test runs, so could be ready for merging.

PR #204, opened 2022-06-27
Enhancement and clarification of provider documentation

mfd2007:
I transfered the documentation to TOML syntax, so it can be directly used as example config.

### DIFF
--- a/docs/csaf_provider.md
+++ b/docs/csaf_provider.md
@@ -32,7 +32,7 @@ An example TOML file with all config options and default values:
 # Let user send password with the request, if set to true the input-field in the web interface will be disappeared.
 #no_passphrase = false
 
-# Validate the uploaded CSAF document against the JSON schema.
+# Skip validation of the uploaded CSAF document against the JSON schema. Default: false.
 #no_validation = false
 
 # Disable the web interface.
--- a/docs/csaf_provider.md
+++ b/docs/csaf_provider.md
@@ -5,33 +5,64 @@ explain how to wire this up with nginx and where the config file lives.
 
 
 ## Provider options
+An example TOML file with all config options and default values:
 
-Following options are supported in the config file:
-
- - password: Authentication password for accessing the CSAF provider.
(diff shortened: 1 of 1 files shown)


bernhardreiter:
Hi @mfd2007,
thanks for the suggestion and the pull-request going along with it. :)
We have not yet decided, if we like this documentation format better, but we will discuss this in the next days. 

(Meanwhile I saw a few minor things, I'll comment on.)
Regards

docs/csaf_provider.md
```diff
@@ -5,33 +5,64 @@ explain how to wire this up with nginx and where the config file lives.
...
+#folder = "/var/www/"
+
+# Specify the web folder
+#web = "/var/www/html"
+
+# Send signature with the request, an additional input-field in the web interface will be shown to let user enter an ascii armored signature.
+#upload_signature = false
+
+# start of the URL where contents shall be accessible from the internet.
+#canonical_url_prefix  = "https://$SERVER_NAME"
```
bernhardreiter, line 30: The default value does not work, if no value is set, it will be configured like shown from the cgi variable SERVER_NAME.

docs/csaf_provider.md
```diff
@@ -5,33 +5,64 @@ explain how to wire this up with nginx and where the config file lives.
...
+
+# Let user send password with the request, if set to true the input-field in the web interface will be disappeared.
+#no_passphrase = false
+
+# Skip validation of the uploaded CSAF document against the JSON schema. Default: false.
```
bernhardreiter, line 35: This one still have a default, like the next. Some does not.

bernhardreiter:
We like the idea, we should place this as docs/example/ file and use markdown-autodocs

JanHoefelmeyer:
I now created https://github.com/csaf-poc/csaf_distribution/pull/236 on the basis of your work. I created a seperate config.toml example file in docs/examples and let the csaf_provider.md reference it instead. Thank you very much for your work.

Issue #82, opened 2022-03-23
provider add config option to require client certificate and password for write access


bernhardreiter:
Wish to have a csaf_provider setup option that a password is required together with a client certificate.

tschmidtb51:
@bernhardreiter You are right that there is not much security gained through a general password.

The use case was that a company use a TLS proxy (and terminates TLS connections there to inspect them). That would make client certificates pretty complicated. Instead of bypassing the TLS proxy for that site, the company installs a TLS certificate at the TLS proxy. However, that would allow anyone from the company to access the web interface and upload CSAF documents. To prevent that they want to use a password for uploading...

Does that make sense?

bernhardreiter:
@tschmidtb51 thanks for describing the use case in more detail.

If the TLS proxy terminates the TLS connection, then there is no client certificate given to the nginx server of the csaf_provider and thus the client certification cannot be used for authentication of the provider. (I have little experience with this setup, so this is  just from reading about such setups.)

From the csaf_provider's point of view, we already support the use case as someone can configure a global password to be acccepted. And the provider would not need to validate and required both password and client certificate.

From the uploader's point of view, we would need to give a client certificate to the proxy and the password to the csaf_provider. This seems to be the case already as the password will be set as extra header `X-CSAF-PROVIDER-AUTH`  and the TLS proxy will use and remove the regular client certificate.

bernhardreiter:
Two remarks: I'd say that the TLS proxy who validates the client certificate should also check which client certificate is allowed to access the csaf_providers cgi-bin interface. We then still need the password, because we had decided that from the providers perspective at least one authentication method must be visible and deployed.

Using a TLS proxy in this way also means that the simple web interface for uploading cannot be used because there is no way to set the header with the password.

tschmidtb51:
@bernhardreiter Sorry, my description seems to be misleading. The use case was meant to have the TLS proxy in the company:

```
 ______________________________       /^^^^^^^^^^\       _______________________________
|  ________        __________  |     /            \     |  _______      ______________  | 
| | client |  => | TLS proxy | | => ( THE INTERNET ) => | | nginx | => | csaf_provider| |
| |________|     |___________| |     \            /     | |_______|    |______________| | 
|______________________________|       ^^^^^^^^^^^      |_______________________________| 
```
[...]

PR #205, opened 2022-06-27
Clarification of authentication configuration

mfd2007:
It should be clarified that the ```password``` by the provider and ```-p, --password=PASSWORD  ```by the uploader is not used/ignored, if client certificates are used. According to the documentation I was expecting that both are used, but that's not the case.


bernhardreiter:
Thanks for the suggestion!
With #82 we will change the behavior, it makes sense to then refit the documentation.

bernhardreiter:
With #82 implemented, there is now an option to control if both client certificate and password are used or not
(in the provider).  Do you think the documentation in the current main is good enough?

Otherwise does it makes sense to change the command line hints to include that it depends on the provider (aka server) configuration?

mfd2007:
I'm fine with the changes in the provider. So my addtional hints to the description of the provider and uploader are not necessary anymore.

bernhardreiter:
@mfd2007 thanks for your feedback!

PR #209, opened 2022-07-01
Config etc

JanHoefelmeyer:
Changes the default location of config.toml from /usr/lib/csaf/config.toml to /etc/csaf/config.toml


cmd/csaf_provider/config.go
```diff
@@ -23,10 +23,11 @@ import (
 const (
 	// The environment name, that contains the path to the config file.
 	configEnv                = "CSAF_CONFIG"
-	configPrefix             = "/usr/lib/csaf"
+	configKeyPrefix          = "/usr/lib/csaf"
```
tschmidtb51, line 26: Is there a reason to have the keys still in `/usr/lib/csaf/`? What does the FHS suggest?

tschmidtb51 (commented):
See below.

cmd/csaf_provider/config.go
bernhardreiter: Without checking I also believe that the openpgp key material is configuration to the provider and therefore should also be under `/etc`. Thanks for the pointer.

bernhardreiter:
@JanHoefelmeyer as comment: the default location at /etc/ for the OpenPGP key material (public and private) is just an example. provider-setup.md has more hints about it. (And need to be adjusted as well.)

s-l-teichmann:
Why not moving all the files to /etc?
Why not simply create a sym link from /usr/where/its/is to /etc?

bernhardreiter:
> Why not moving all the files to /etc?

We should do this for all configuration files. 

> Why not simply create a sym link from /usr/where/its/is to /etc?

Not adhereing to FHS in my understanding.

Issue #77, opened 2022-03-02
Provide ROLIE as prefered default.


bernhardreiter:
For a correct CSAF Trusted Provider, only one method of providing several things is needed.

It would be nice to have an option to only publish as ROLIE and to have this option being activated by default.
Because ROLIE is considered the preferred way of publishing via CSAF-2.0 and by using it only, a Trusted Provider sets a good example for others.

#### Maybe out of scope for 1.0

bernhardreiter:
(is contracted and planned now)

PR #210, opened 2022-07-01
Only create/update index.txt, changes.csv when configured.

s-l-teichmann:
Solves #77 

Maybe TODO: Don't allow directory listing of folders containing advisories?


bernhardreiter:
Running the integration test via github action brings (https://github.com/csaf-poc/csaf_distribution/actions/runs/2616532451) : good index and changes found? Something is amiss.

Issue #73, opened 2022-03-02
Allow for "listing only" entries in the CSAF Aggregator configuration


bernhardreiter:
CSAF Aggregators must at least do two mirrors, but can also have entries that are just listed, see
https://github.com/oasis-tcs/csaf/issues/470

### technical consideration
A config option per provider could handle this.

bernhardreiter:
After the creation of this issue, the CSAF specification was updated, by https://github.com/oasis-tcs/csaf/pull/472/files in this point.

JanHoefelmeyer:
To specify the problem to my understanding: Aggregators can be of the categories "aggregator", which is a mirror, and "lister", which only lists the data. 

https://docs.oasis-open.org/csaf/csaf/v2.0/csaf-v2.0.html Example 138 shows a lister, example 139 an aggregator.

There need to be 2+ entries in the category aggregator, and 0+ entries in the category lister for the distributing party to qualify as "CSAF aggregator" 

This means currently we're not checking whether we have 2+ aggregators but only whether we got 2+ providers, which would falsely allow one aggregator and one lister or two listers to wrongly satisfy the requirement when they should fail, which needs to be fixed.

JanHoefelmeyer:
@tschmidtb51 , if the aggregator.toml contains at least 2 providers, but less than 2 aggregators (mirrors), is the aggregator supposed to throw an error?

tschmidtb51:
Yes, as this violates the third point:
> lists a mirror for at least two disjoint issuing parties pointing to a domain under its own control.

PR #214, opened 2022-07-08
Provider list mirror

JanHoefelmeyer:
Solves https://github.com/csaf-poc/csaf_distribution/issues/73


cmd/csaf_aggregator/config.go
```diff
@@ -85,6 +87,34 @@ func (p *provider) writeIndices(c *config) bool {
 	return c.WriteIndices
 }
 
+func (p *provider) runAsMirror(c *config) bool {
+	if p.AggregatoryCategory != nil {
+		return *p.AggregatoryCategory == csaf.AggregatorAggregator
+	}
+	err := errors.New("Couldn't parse aggregator.toml: Provider has no category.")
+	log.Fatal(err)
```
s-l-teichmann, line 95: This will let the  program die!

cmd/csaf_aggregator/config.go
```diff
@@ -168,6 +198,13 @@ func (c *config) checkProviders() error {
 	return nil
 }
 
+func (c *config) allowMisconfigure() bool {
```
s-l-teichmann, line 201: Remove this pointless method.

cmd/csaf_aggregator/full.go
```diff
@@ -113,13 +117,23 @@ func (p *processor) full() error {
 	for i := 1; i <= p.cfg.Workers; i++ {
 		wg.Add(1)
 		w := newWorker(i, p)
-		go w.fullWork(&wg, doWork, queue)
+		//go w.fullWork(&wg, doWork, queue)
```
s-l-teichmann, line 120: Remove this line.

cmd/csaf_aggregator/full.go
```diff
@@ -99,11 +93,21 @@ func (p *processor) full() error {
 			}()
 		}
 
-		doWork = (*worker).mirror
 		log.Println("Running in aggregator mode")
+		if !p.cfg.hasTwoMirrors() {
+			if !p.cfg.allowMisconfigure(){
+				err := errors.New("At least 2 Providers need to be mirrors.")
+    			        log.Fatal(err)
```
s-l-teichmann, line 100: This will let the program die.

cmd/csaf_aggregator/config.go
```diff
@@ -85,6 +87,34 @@ func (p *provider) writeIndices(c *config) bool {
...
+		}
+	}
+	return false
+}
+
+func (c *config) hasTwoMirrors() bool {
+	var numberMirrors = 0
+        for _, p := range c.Providers {
+                if p.AggregatoryCategory != nil && *p.AggregatoryCategory == csaf.AggregatorAggregator {
+                        numberMirrors += 1
```
s-l-teichmann, line 112: if numberMirrors++; numberMirrows > 1 {
    return true
}

cmd/csaf_aggregator/config.go
```diff
@@ -85,6 +87,34 @@ func (p *provider) writeIndices(c *config) bool {
...
+                }
+        }
+        return numberMirrors > 1
```
s-l-teichmann, line 115: return false

cmd/csaf_aggregator/config.go
```diff
@@ -85,6 +87,34 @@ func (p *provider) writeIndices(c *config) bool {
...
+	for _, p := range c.Providers {
+		if p.AggregatoryCategory != nil && *p.AggregatoryCategory == csaf.AggregatorAggregator {
+			return true
+		}
+	}
+	return false
+}
+
+func (c *config) hasTwoMirrors() bool {
+	var numberMirrors = 0
```
s-l-teichmann, line 109: numberMirrows := 0
or
var numberMirrows int

s-l-teichmann:
@JanHoefelmeyer Before you put any more time into this we need to talk about this.

JanHoefelmeyer:
Tested commit 7cf06b0 and found it working as intended. Should be able to be merged after review.

cmd/csaf_aggregator/config.go
```diff
@@ -36,12 +36,13 @@ type provider struct {
...
-	Categories *[]string `toml:"categories"`
+	Rate         *float64  `toml:"rate"`
+	Insecure     *bool     `toml:"insecure"`
+	WriteIndices *bool     `toml:"write_indices"`
+	Categories   *[]string `toml:"categories"`
 	// ServiceDocument incidates if we should create a service.json document.
-	ServiceDocument *bool `toml:"create_service_document"`
-	WriteIndices    *bool `toml:"write_indices"`
+	ServiceDocument     *bool                    `toml:"create_service_document"`
+	AggregatoryCategory *csaf.AggregatorCategory `toml:"category"`
```
bernhardreiter, line 45: There seems to be a potential for confusion as a provider strictly does not have an AggregatoryCategory, because a provider is not an aggregator. I suggest to use a different name and type for this configuration option, what about bool and something about to_be_mirrored or so?

docs/csaf_aggregator.md
```diff
@@ -134,12 +135,22 @@ insecure = true
   domain = "localhost"
 #  rate = 1.5
 #  insecure = true
+   category = "lister"
```
bernhardreiter, line 138: changes here are in the wrong file, it gets inserted from docs/examples/aggregator.toml

docs/csaf_aggregator.md
```diff
@@ -101,6 +101,7 @@ domain
 rate
 insecure
 write_indices
+category
```
bernhardreiter, line 104: Needs a documentation (which values can be used and what is there effect)

docs/csaf_aggregator.md
bernhardreiter: (Okay, the changes are also in  docs/examples/aggregator.toml, so it can stay in there, would just be unnecessary as it gets replaced.)

bernhardreiter (changes requested):
My suggestion is to change the "category" option to a bool (as indicated in the comment) and then document it

bernhardreiter:
Spoke with @s-l-teichmann , we see several way how to shape the configuration (detailed below).

@tschmidtb51 to you think, users would prefer?


### Way 1: use `category` to show that it affects the category of the aggregator

```toml
[aggregator]
  category = "aggregator"
#...

[[providers]]  
    name = "A"
    category = "aggregator"
#...

[[providers]]
    name = "B"
    category = "aggregator"
#...

[[providers]]
    name = "C"
    category = "lister"
#...
```

Cons:
* Potential confusion because C's `lister` does not change the category of the aggregator.
* Config writers need to know that they need to use `aggregator` mark an entry for mirroring.


### Way 2: use an option name which explains what shall happen to the provider entry

```toml
[aggregator]
#  category = "aggregator" # not stricely necessary as it can be derived
#...

[[providers]]  
    name = "A"
    to_be_mirrored = true
#...

[[providers]]
    name = "B"
    to_be_mirrored = true
#...

[[providers]]
    name = "C"
    to_be_mirrored = false
#...
```

Cons:
* If more possible kinds of provider entries appear, a boolean is not enough. Would need a string.
* By default aggregator.category could be set in the config file and we would need to actively check
that it does not conflict with the entries.


### Way 3 use `role` instead of `category`?
Con: `role` is also used quite often, so may not clarify things well enough.

bernhardreiter:
Talked to @tschmidtb51, he prefers to have a safeguard against someone who by accident write a mirror entry into a setup which should be a lister. The boolean from way 2 also looks like an improvement. However to be more consistent with other configurations for csaf_distribution, we shall do a way 1 like:

```
[aggregator]
  category = "aggregator"
#...

[[providers]]  
    name = "A"
#...

[[providers]]
    name = "B"
    #category = "aggregator"   not necessary, but possible because default
#...

[[providers]]
    name = "C"
    category = "lister"
#...
```

and 
```
[aggregator]
  category = "lister"
#...

[[providers]]  
    name = "A"
#...

[[providers]]
    name = "B"
    # category = "lister" not necessary, but possible because default
```

We'll test accordingly. I believe the current implement should already enable this.

bernhardreiter (approved):

bernhardreiter:
Fixed the small problem and added documentation for the improved was of configuration.

PR #224, opened 2022-07-14
Remove additional slash for index.txt and changes.csv

tschmidtb51:
fixes #223

PR #226, opened 2022-07-14
Joinpath

s-l-teichmann:
Fixes #224

### DIFF
--- a/util/joinpath.go
+++ b/util/joinpath.go
@@ -1,0 +1,303 @@
+// This file is Free Software under the MIT License
+// without warranty, see README.md and LICENSES/MIT.txt for details.
+//
+// SPDX-License-Identifier: MIT
+//
+// SPDX-FileCopyrightText: 2022 German Federal Office for Information Security (BSI) <https://www.bsi.bund.de>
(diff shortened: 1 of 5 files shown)


util/joinpath.go
```diff
@@ -0,0 +1,306 @@
...
+// without warranty, see README.md and LICENSES/MIT.txt for details.
+//
+// SPDX-License-Identifier: MIT
+//
+// SPDX-FileCopyrightText: 2022 German Federal Office for Information Security (BSI) <https://www.bsi.bund.de>
+// Software-Engineering: 2022 Intevation GmbH <https://intevation.de>
+
+//go:build !go1.19
+
+// Copyright 2009 The Go Authors. All rights reserved.
```
bernhardreiter, line 11: We need to properly attribute the file contents, this means the right REUSE headers, which includes adding the LICENSE to our LICENSES directory and a clear pointer to where we've gotten this file in particular.

bernhardreiter: @s-l-teichmann can you add the precise link where you've gotten the file contents from.

s-l-teichmann: Should be fixed by now.

Issue #207, opened 2022-06-30
Handle non-json errors in uploader if config.toml is incorrect


JanHoefelmeyer:
Version: 0.9.3-11-g3a3ef7a

Using invalid config.toml files (in /usr/lib/csaf/config.toml) returns ```HTTPS upload failed: 500 Internal Server Error``` instead of the expected error messages when using the uploader.

Reproduce: 
1. open the config.toml file and either enter unknown values (e.g. wurst = "wurst") or invalid values (e.g. tlps = [1,2])
2. run the uploader with any valid file (tested with  ./docs/scripts/csaf_examples/"rhsa-2021_5186.json")

Expectation: 
Response is ```Config error: Couldn't parse ["wurst"] from config.toml``` (unknown value) or ```Config error: invalid config TLP value: 1.200000``` (invalid value)

Result:
```HTTPS upload failed: 500 Internal Server Error```

Likely due to error:
```invalid character 'C' looking for beginning of value```
thrown in 
    ````if err := json.NewDecoder(resp.Body).Decode(&result); err != nil {
                return err
        }````
in csaf_uploader, since the error messages are strings and not json.

bernhardreiter:
related to #94

bernhardreiter:
Server side has been improved meanwhile.

JanHoefelmeyer:
Improved through https://github.com/csaf-poc/csaf_distribution/pull/247

PR #247, opened 2022-07-21
Handle none JSON replies to uploader calls as errors.

JanHoefelmeyer:
Fixes not-yet-fixed of https://github.com/csaf-poc/csaf_distribution/issues/207

### DIFF
--- a/cmd/csaf_uploader/main.go
+++ b/cmd/csaf_uploader/main.go
@@ -15,11 +15,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"log"
 	"mime/multipart"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


bernhardreiter:
The error message to look into the logs should come from the server.

The client (which is the uploader) should then display it. 
(By the logic: Do we expect a json as feedback: Yes, then parse it and act accordingly. If parsing fails, display the start of the plan text body. If we do not expect a json, we could display the plain text body right away.)

s-l-teichmann:
If have added a check  for content-type "application/json" before parsing it.
This should handle more gracefully.
Simply swallowing the JSON decoder reason was not right. For example the server could have started with a
correct JSON response abut then the connection may become broken some how.
This situation should still be detectable.

PR #248, opened 2022-07-21
Improve integration test scripts

bernhardreiter:
* Add a simple aggegator web server config.
  * Add a downloader test using a direct url from the aggregator.


bernhardreiter:
The results from https://github.com/csaf-poc/csaf_distribution/runs/7451386738 look good to me regarding the topic of this pull request, they are so good, that they find a defect. ;) (#249)

PR #262, opened 2022-07-26
Change default port for secvisogram

bernhardreiter:
* Following change from
   https://github.com/secvisogram/csaf-validator-service/pull/14

### DIFF
--- a/docs/scripts/setupProviderForITest.sh
+++ b/docs/scripts/setupProviderForITest.sh
@@ -116,7 +116,7 @@ set +e
 echo
 echo === Try to reach the validator service ten times, up to nine fails are ok
 for ((i = 1; i <= 10; i++)); do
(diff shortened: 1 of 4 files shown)


docs/scripts/setupProviderForITest.sh
```diff
@@ -116,7 +116,7 @@ set +e
 echo
 echo === Try to reach the validator service ten times, up to nine fails are ok
 for ((i = 1; i <= 10; i++)); do
- if [ $(curl -IL http://localhost:3000/api/v1/tests | grep -c HTTP ) != "0" ]; then
+ if [ $(curl -IL http://localhost:8082/api/v1/tests | grep -c HTTP ) != "0" ]; then
  break
  fi
```
s-l-teichmann, line 121: We should stop the script execution if we can't start the service.

PR #264, opened 2022-07-26
Load client certs early to detect misconfiguration

s-l-teichmann:
Do not die with `log.Fatalf` in the middle of checking when client certs are configured badly.
Check this right at the start.

### DIFF
--- a/cmd/csaf_checker/main.go
+++ b/cmd/csaf_checker/main.go
@@ -10,6 +10,7 @@ package main
 
 import (
 	"bufio"
+	"crypto/tls"
 	_ "embed" // Used for embedding.
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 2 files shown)


bernhardreiter:
@s-l-teichmann what kind of missconfiguration would trigger this?

s-l-teichmann:
> @s-l-teichmann what kind of missconfiguration would trigger this?

Something not being a valid certificate file like:
```
./csaf_checker --client-cert=/etc/passwd --client-key=/etc/passwd  
2022/07/26 17:33:25 error: tls: failed to find any PEM data in certificate input
```

bernhardreiter:
In my tests this already fails early, because the current check is triggered by p.httpClient()
 which happens to be called by checkProviderMetadata(), which is the first test.
Is the change more about the right place where to check and a premature death cannot be triggered?

s-l-teichmann:
> In my tests this already fails early, because the current check is triggered by p.httpClient() which happens to be called by checkProviderMetadata(), which is the first test. Is the change more about the right place where to check and a premature death cannot be triggered?

What ever you mean by premature death.
This PR loads the configured certs before the checking machine is started ... and yes, this the right place to do this.

s-l-teichmann:
`p.httpClient()` should not die ... if this creates an error, it should be properly handled with an error code path.

Issue #271, opened 2022-08-02
Interim mode of aggregator: Problem with filename and not working


mfd2007:
Version: 0.9.5

There is smix between the used files:
In the normal aggregator mode for interim CSAF files the file `interim.csv` is created:
https://github.com/csaf-poc/csaf_distribution/blob/1b0e5b4238515cbbba3b1cae8b313a3cf068b87b/cmd/csaf_aggregator/indices.go#L47

In interim-mode the file `interims.csv` will be loaded. 
https://github.com/csaf-poc/csaf_distribution/blob/78d8b89aca3bc226d07ba4aea513471a9c4a350d/cmd/csaf_aggregator/interim.go#L179

As a result, no interim documents are downloaded.

s-l-teichmann:
Doh! Shame on me. PR #274 should fix this.

s-l-teichmann:
PR #275 addresses this issue + some other weirdness I've found on the way.

PR #274, opened 2022-08-02
Fixed spelling of interims.csv

s-l-teichmann:
Solves #271


s-l-teichmann:
During testing I run in more issues.
These issues including #271 are addressed in PR #275

PR #315, opened 2023-01-26
Improve setup script für Ubuntu systems

bernhardreiter:
* Make sure that we do not have an old go installation around, because otherwise it maybe broken by untarring the newer one. As documented on https://go.dev/doc/install .


s-l-teichmann (approved):
LGTM

JanHoefelmeyer (approved):
Works on my end as well

PR #313, opened 2023-01-25
Add remote validator support to checker

s-l-teichmann:
This PR adds support for remote validation to the checker.

This partly resolves #300

PR #321, opened 2023-01-26
Improve checker help file

### LINKED ISSUE #320: Improve csaf_checker docs
The [`csaf_checker`](https://github.com/csaf-poc/csaf_distribution/blob/main/docs/csaf_checker.md) and the `csaf_checker -h` do not announce that a `domain` must be given. This should be added.

tschmidtb51:
* fixes #320
* Add "domain" to the usage and documentation.

### DIFF
--- a/docs/csaf_checker.md
+++ b/docs/csaf_checker.md
@@ -3,7 +3,7 @@
 ### Usage
 
 ```
-csaf_checker [OPTIONS]
+csaf_checker [OPTIONS] domain...
 
 Application Options:
   -o, --output=REPORT-FILE       File name of the generated report
--- a/cmd/csaf_checker/main.go
+++ b/cmd/csaf_checker/main.go
@@ -157,7 +157,9 @@ func buildReporters() []reporter {
 func main() {
 	opts := new(options)
 
-	domains, err := flags.Parse(opts)
+	parser := flags.NewParser(opts, flags.Default)
+	parser.Usage = "[OPTIONS] domain..."
+	domains, err := parser.Parse()
 	errCheck(err)
 
 	if opts.Version {


docs/csaf_checker.md
```diff
@@ -3,7 +3,7 @@
 ### Usage
 
 ```
-csaf_checker [OPTIONS]
+csaf_checker [OPTIONS] domain...
```
tschmidtb51, line 6: Is the `...` the common notation for "you can have multiple domains"?

s-l-teichmann: I think so ... `LANG=C cat --help | head -1` gives:
```
Usage: cat [OPTION]... [FILE]...
```

Maybe  we should use `[OPTION]... domain...` ?

tschmidtb51: Tough question. It looks like the go default is just `[OPTIONS]`. But I like to Linux version as it is

tschmidtb51: BTW: Whatever we decide - I'm going to implement it...

bernhardreiter: Usually it is good to follow the 
 * https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html  GNU Coding standards
which point to the POSIX standards, they extend.
 * https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/
 
Looking there and at a few examples, it seems that several things are implemented and possible,
both with
```
  utility_name [options][operands]
```
and with dots like `...` 
```
chgrp -R [-H|-L|-P] group file...
```
and
```
utility_name [-g option_argument]...[operand...]
```

As the checker has many options, ```[options]``` or ```[OPTIONS]``` seems best practice
as we cannot list all options in there and the dots are used to indicate that one options can be used
several times (in posix). As we can name the operand specifically, either `domain` or a better term
we should list them and then use the dots, so I propose to use

```
csaf_checker [OPTIONS] domain...
```
(it is shorter and seems closer to posix.)

bernhardreiter (approved):

tschmidtb51:
Close in favor of #313

PR #322, opened 2023-01-27
Uploader: Log the filename too, if an upload fails. 

### LINKED ISSUE #292: Improve error output
Currently, it could sometimes be complicated to find out for which file the local schema check failed when doing a batch upload. We should add the filename to the error message.

s-l-teichmann:
Resolves #292

### DIFF
--- a/cmd/csaf_uploader/main.go
+++ b/cmd/csaf_uploader/main.go
@@ -435,6 +435,8 @@ func main() {
 	}
 
 	for _, arg := range args {
-		check(p.process(arg))
+		if err := p.process(arg); err != nil {
+			log.Fatalf("error: processing %q failed: %v\n", arg, err)
+		}
 	}
 }


tschmidtb51 (approved):
LGTM

tschmidtb51:
LGTM and works.

bernhardreiter (approved):
LGTM

cmd/csaf_uploader/main.go
```diff
@@ -435,6 +435,8 @@ func main() {
 	}
 
 	for _, arg := range args {
-		check(p.process(arg))
+		if err := p.process(arg); err != nil {
```
bernhardreiter, line 438: This changes the semantics a bit, because `check()` will use os.exit(0) in some cases, the new code doesn't. @s-l-teichmann is this intentional?

bernhardreiter (changes requested):

cmd/csaf_uploader/main.go
s-l-teichmann: `check`  only does `os.exit(0)` if the command line help was written. This case does not apply here. All errors come from the processing.

bernhardreiter (approved):
LGTM

PR #328, opened 2023-01-31
Enforce `application/json` when uploading advisories to provider.

### LINKED ISSUE #148: Reject other than application/json content-types
When trying to upload an advisory, the `csaf_provider` should reject documents which are not declared as `application/json` in the HTTP content-type header.

s-l-teichmann:
Resolves #148

### DIFF
--- a/cmd/csaf_provider/actions.go
+++ b/cmd/csaf_provider/actions.go
@@ -42,6 +42,11 @@ func (c *controller) loadCSAF(r *http.Request) (string, []byte, error) {
 	}
 	defer file.Close()
 
+	// We reject everything which is not announced as JSON.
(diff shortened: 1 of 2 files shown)


tschmidtb51 (approved):
LGTM and works.

bernhardreiter (dismissed):
LGETM
(When reading the code, it would have been helpful to note even more clear that the escaper and the CreateFromFile is copying the main code from https://cs.opensource.google/go/go/+/refs/tags/go1.20:src/mime/multipart/writer.go;l=140 which is not entirely obvious as the function is called _From_ while the original is called _Form_ .)

bernhardreiter (approved):
LGTM

bernhard-herzog (approved):
LGTM.

PR #329, opened 2023-02-01
Warn in checker and downloader if advisories are delivered as none JSON

### LINKED ISSUE #311: report if documents do not have expected content-type application/json
The checker and downloader should warn if a document is required to be type application/json, but the webserver does not use this as a content type in the HTTP headers.

It would become a reject / error, if the CSAF standard demands it. (And open question.)

s-l-teichmann:
Resolves #311

### DIFF
--- a/cmd/csaf_downloader/downloader.go
+++ b/cmd/csaf_downloader/downloader.go
@@ -281,6 +281,13 @@ func (d *downloader) downloadFiles(label csaf.TLPLabel, files []csaf.AdvisoryFil
 			continue
 		}
 
+		// Warn if we do not get JSON.
+		if ct := resp.Header.Get("Content-Type"); ct != "application/json" {
+			log.Printf(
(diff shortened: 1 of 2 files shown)


cmd/csaf_downloader/downloader.go
```diff
@@ -281,6 +281,13 @@ func (d *downloader) downloadFiles(label csaf.TLPLabel, files []csaf.AdvisoryFil
 			continue
 		}
 
+		// Warn if we do not get JSON.
+		if ct := resp.Header.Get("Content-Type"); ct != "application/json" {
+			log.Printf(
+				"WARN: The content type of %s should be 'application/json but is '%s'\n",
```
JanHoefelmeyer, line 287: The quotes around application/json are not closed. ('application/json -> 'application/json' )

tschmidtb51 (approved):
LGTM and works as expected.

Issue #302, opened 2022-11-14
Add a check for requirement 2


tschmidtb51:
We should add a check for [requirement 2](https://docs.oasis-open.org/csaf/csaf/v2.0/csaf-v2.0.html#712-requirement-2-filename) to the `csaf_checker`.

This might also be a question for the other programs...

bernhardreiter:
The downloader shall reject non-compliant filenames.

s-l-teichmann:
PR #330 adds this to the checker.

State of the other tools:
* The uploader checks the schema optionally. This is okay IMHO because:
* The validator checks the schema and the remote validator if configured.
* The provider checks the the schema and uses the remote validator if configured.
* The downloader checks the schema and the remote validator if configured,
* The aggregator checks the schema and the remote validator if configured.

Issue #301, opened 2022-11-14
Add check for requirement 1


tschmidtb51:
We should add a check for [requirement 1](https://docs.oasis-open.org/csaf/csaf/v2.0/csaf-v2.0.html#711-requirement-1-valid-csaf-document) to the `csaf_checker`. This requires #300.

bernhardreiter:
Must be seen in the reporting.
 * schema check is done already.
 *  if remote_validator is not available, note this in reporting (`no mandatory tests done` or so)

s-l-teichmann:
PR #330 implements this.

The check results where formerly reported in the Integrity section.
Now they are added to the new "Valid CSAF documents" one.

PR #330, opened 2023-02-01
Add new requirements 1 and 2 to report

s-l-teichmann:
Solves #301 
Solves #302 

This PR add two new sections to the report generated by the checker:

1. Valid CSAF document
2. Filename

If no remote validator is configured a respective note  is added to section 1.
Sections 1 and 2 lists the URLs which do not pass the respective tests.
It is noted if there were no tests performed.

### DIFF
--- a/cmd/csaf_checker/processor.go
+++ b/cmd/csaf_checker/processor.go
@@ -53,6 +53,8 @@ type processor struct {
 	pmd            any
 	keys           []*crypto.KeyRing
 
+	invalidAdvisories    topicMessages
+	badFilenames         topicMessages
 	badIntegrities       topicMessages
 	badPGPs              topicMessages
 	badSignatures        topicMessages
@@ -203,6 +205,8 @@ func (p *processor) clean() {
 	p.pmd = nil
 	p.keys = nil
 
+	p.invalidAdvisories.reset()
+	p.badFilenames.reset()
 	p.badIntegrities.reset()
 	p.badPGPs.reset()
 	p.badSignatures.reset()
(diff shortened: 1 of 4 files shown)


cmd/csaf_checker/reporters.go
```diff
@@ -43,6 +45,33 @@ func (bc *baseReporter) requirement(domain *Domain) *Requirement {
...
+	if !p.invalidAdvisories.used() {
+		req.message(InfoType, "No validations performed")
+	} else if len(p.invalidAdvisories) == 0 {
+		req.message(InfoType, "All advisories validated fine.")
+	} else {
+		req.Append(p.invalidAdvisories)
+	}
+}
+
+// report reposrts if there where any bad filename.
```
tschmidtb51, line 63:
```suggestion
// report reports if there where any bad filename.

**Review by tschmidtb51 — DISMISSED** [review]
Only a minor spelling mistake. Other than that: LGTM and works

**Review by tschmidtb51 — APPROVED** [review]
LGTM and works

File: cmd/csaf_checker/processor.go [hunk 2]
```diff
@@ -441,6 +445,12 @@ func (p *processor) integrity(
 		}
 		p.checkTLS(u)
 
+		// Check if the filename is confirming.
```
bernhardreiter, line 448: In my understanding _confirming_ is not the right word. It means the action of confirming something, not the item is confirmed itself. Maybe ""conforming"?

cmd/csaf_checker/reporters.go
```diff
@@ -43,6 +45,33 @@ func (bc *baseReporter) requirement(domain *Domain) *Requirement {
...
+	}
+}
+
+// report reports if there where any bad filename.
+func (r *filenameReporter) report(p *processor, domain *Domain) {
+	req := r.requirement(domain)
+	if !p.badFilenames.used() {
+		req.message(InfoType, "No filenames checked for conformance")
+	} else if len(p.badFilenames) == 0 {
+		req.message(InfoType, "All found filenames are confirming.")
```
bernhardreiter, line 69: see above _conforming_ versus  being _conforming to_

bernhardreiter (changes requested):
Clarify the message for invalid filesnames.

cmd/csaf_checker/processor.go
bernhardreiter: But this does not affect the funcationality.

bernhardreiter (approved):
LGETM

Issue #324, opened 2023-01-30
Add option to see  results of remote validator


tschmidtb51:
Please add an option to see the results of the remote validator (as one could use that to run optional or informative tests).

_Originally posted by @tschmidtb51 in https://github.com/csaf-poc/csaf_distribution/pull/323#discussion_r1090236695_

tschmidtb51:
The [remote validation code](https://github.com/csaf-poc/csaf_distribution/blob/0c398199303a2bb0ddc76a5046933735789ef115/csaf/remotevalidation.go#L97) currently does not handle details of the validation.

To do so this code has to be extended.

_Originally posted by @s-l-teichmann in https://github.com/csaf-poc/csaf_distribution/pull/323#discussion_r1090619138_

bernhardreiter:
Accepted would be a display of results in JSON Format.

@tschmidtb51 to post an example

tschmidtb51:
A request with
```
curl -X 'POST' \
  'http://localhost:8082/api/v1/validate' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "tests": [
    {
      "name": "basic",
      "type": "preset"
    }
  ],
  "document": {
    "additionalProp1": {}
  }
}'
```
results in:
```
{
  "isValid": false,
  "tests": [
    {
      "errors": [
        {
          "instancePath": "",
          "message": "must have required property 'document'"
        },
        {
          "instancePath": "",
          "message": "must NOT have additional properties"
        }
      ],
      "infos": [],
      "warnings": [],
      "isValid": false,
      "name": "csaf_2_0_strict"
    },
    {
      "errors": [],
      "infos": [],
      "warnings": [],
      "isValid": true,
      "name": "mandatoryTest_6_1_1"
    },
    {
      "errors": [],
      "infos": [],
      "warnings": [],
      "isValid": true,
      "name": "mandatoryTest_6_1_10"
    },
    {
      "errors": [],
      "infos": [],
      "warnings": [],
      "isValid": true,
      "name": "mandatoryTest_6_1_11"
    },
    {
      "errors": [],
      "infos": [],
      "warnings": [],
      "isValid": true,
      "name": "mandatoryTest_6_1_12"
    },
    {
      "errors": [],
      "infos": [],
      "warnings": [],
      "isValid": true,
      "name": "mandatoryTest_6_1_13"
    },
    {
      "errors": [],
      "infos": [],
      "warnings": [],
[...]

PR #347, opened 2023-03-02
Remote validator output

JanHoefelmeyer:
Solves https://github.com/csaf-poc/csaf_distribution/issues/324 by adding the --verbose flag to the csaf_validator.

Calling the validator with the verbose flag but without a remote validator currently has no effect. (The validator behaves as if verbose was not called)

### DIFF
--- a/cmd/csaf_validator/main.go
+++ b/cmd/csaf_validator/main.go
@@ -26,7 +26,7 @@ type options struct {
 	RemoteValidator        string   `long:"validator" description:"URL to validate documents remotely" value-name:"URL"`
 	RemoteValidatorCache   string   `long:"validatorcache" description:"FILE to cache remote validations" value-name:"FILE"`
 	RemoteValidatorPresets []string `long:"validatorpreset" description:"One or more presets to validate remotely" default:"mandatory"`
-	Output                 string   `short:"o" long:"output" description:"If a remote validator was used, display the results in JSON format"`
+	Output                 string   `short:"o" long:"output" description:"If a remote validator was used, display AMOUNT ('all', 'important' or 'short') results" value-name:"AMOUNT"`
 }
 
 // ShortTest is the result of the remote tests
--- a/cmd/csaf_validator/main.go
+++ b/cmd/csaf_validator/main.go
@@ -88,6 +88,8 @@ func run(opts *options, files []string) error {
 		printResult = printShort
 	case "important":
 		printResult = printImportant
+	case "":
+		printResult = noPrint
 	default:
 		return fmt.Errorf("unknown output amount %q", opts.Output)
 	}
@@ -140,6 +142,8 @@ func run(opts *options, files []string) error {
 
 	return nil
 }
+
+func noPrint(*csaf.RemoteValidationResult) error { return nil }
 
 func printShort(rvr *csaf.RemoteValidationResult) error {
 	short := ShortTest{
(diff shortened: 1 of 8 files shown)


cmd/csaf_validator/main.go
```diff
@@ -26,6 +26,7 @@ type options struct {
 	RemoteValidator        string   `long:"validator" description:"URL to validate documents remotely" value-name:"URL"`
 	RemoteValidatorCache   string   `long:"validatorcache" description:"FILE to cache remote validations" value-name:"FILE"`
 	RemoteValidatorPresets []string `long:"validatorpreset" description:"One or more presets to validate remotely" default:"mandatory"`
+	Verbose                bool     `long:"verbose" description:"If a remote validator was used, display the results in JSON format"`
```
tschmidtb51, line 29: In the other programs `-v` is additionally short for  `--verbose`

csaf/remotevalidation.go
```diff
@@ -55,9 +56,37 @@ type outDocument struct {
...
 }
 
+// rtest is the result of the remote tests
+// recieved by the remote validation service.
+type rtest struct {
+	Valid   bool        `json:"isValid"`
+	Warning []rwarnings `json:"warnings"`
+	Error   []rerrors   `json:"errors"`
+	Info    []rinfos    `json:"infos"`
+}
```
tschmidtb51, line 66: The test name should be returned as well `name`. Otherwise, it is hard to figure out which test is which.
```suggestion
        Name    string    `json:"name"`
}

**Review by tschmidtb51 — CHANGES_REQUESTED** [review]
Some minor changes...

File: csaf/remotevalidation.go (hunk 2 above)
Comment (tschmidtb51) [reply csaf/remotevalidation.go:66]: Please also sort keys (errors, warning, infos) it to match https://github.com/csaf-poc/csaf_distribution/issues/324#issuecomment-1421665324

**Comment by tschmidtb51** [comment]
Given #348, I'm not 100% sure whether `--verbose` is the right keyword...

File: cmd/csaf_validator/main.go (hunk 1 above)
Comment (JanHoefelmeyer) [reply cmd/csaf_validator/main.go]: Along with renaming verbose to output due to the concerns in https://github.com/csaf-poc/csaf_distribution/pull/347#issuecomment-1452470345, the -o flag has now been added

File: csaf/remotevalidation.go (hunk 2 above)
Comment (JanHoefelmeyer) [reply csaf/remotevalidation.go:67]: Has now been done

**Review by s-l-teichmann — CHANGES_REQUESTED** [review]
There are many things I dont like about this.
@JanHoefelmeyer  we need to talk about this.
We should set this back to draft.

File: csaf/remotevalidation.go (hunk 2 above)
Comment (tschmidtb51) [reply csaf/remotevalidation.go:67]: Thanks for pointing that out - I didn't check the order in the comment. The order should match: https://github.com/csaf-poc/csaf_distribution/issues/348#issue-1607401382

**Review by s-l-teichmann — DISMISSED** [review]
LGTM and works.

**Review by s-l-teichmann — APPROVED** [review]
Still LGTM and works.

**Review by tschmidtb51 — APPROVED** [review]
Comment on the Error message - other than that: LGTM

File: cmd/csaf_validator/main.go [hunk 3]
```diff
@@ -95,13 +111,17 @@ func run(opts *options, files []string) error {
...
 		if validator != nil {
-			validate, err := validator.Validate(doc)
+			rvr, err := validator.Validate(doc)
 			if err != nil {
 				return fmt.Errorf("remote validation of %q failed: %w",
 					file, err)
 			}
+			if err := printResult(rvr); err != nil {
+				return fmt.Errorf("remote validation of %q failed: %w",
+					file, err)
```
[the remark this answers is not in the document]
s-l-teichmann: As there where actually no errors raised on this code path I have removed the error handling in this spot completely. This also gets rid of the problem of having a better error message. ;-)

s-l-teichmann (approved):
LGTM

tschmidtb51 (approved):
LGTM

Issue #345, opened 2023-02-28
Updating some libraries failed integration tests.


s-l-teichmann:
Updating the libraries

* github.com/PuerkitoBio/goquery
* github.com/santhosh-tekuri/jsonschema/v5

to their latest versions leads to  failing the integration test.
The reason is currently unknown and needs further investigations.

For PR #343 I left these libaries at at their known working versions.

s-l-teichmann:
In the [up-exp](https://github.com/csaf-poc/csaf_distribution/tree/up-exp)  branch I have upgrade the jsonschema lib. The change in the library are quiet trivial and should not result in failing the integration test. Maybe our integration test is broken? Issue #346 sings this song. Further investigation needed.

PR #349, opened 2023-03-07
Upgrade jsonschema library to 5.2.0

s-l-teichmann:
The jsonschema library introduced a subtle change when compiling schemas with regard of handling external sources. It may be a bug.

The resulting problem can be circumvented by simplifying the usage of
the schema compiler. 

Partly solves #345


JanHoefelmeyer (approved):
Works for me

.github/workflows/itest.yml
```diff
@@ -11,11 +11,17 @@ jobs:
         with:
           go-version: 1.19.1
 
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
+
```
tschmidtb51, line 18: Do we need the setup node action here as we install Node.js manually in https://github.com/csaf-poc/csaf_distribution/pull/349/files#diff-9e14b99a1dd4003f9986a20557f9e3ddbd8d071fc0d62d3e3eec7bf5bf2c0244

tschmidtb51 (commented):
A small comment. I think having the part in the script helps for easy local tests...

.github/workflows/itest.yml
tschmidtb51: Flagging @s-l-teichmann to answer the question...

s-l-teichmann: This was contributed by @bernhardreiter to make the integration test work again. Bernhard, what's the deal?

bernhardreiter: The setup is improved to follow the original pattern:
 * .github/workflows/itest.yml  uses the github actions, which allow caching on the github runner's side.
 * docs/scripts/prepareUbuntuInstanceForITests.sh is used to prepare a naked Ubuntu system once. And is not part of the testing script itself. 
 
This allows the setup steps to be done only once and not per test run on a VM. It makes both the automatic and the manual tests faster (and saves a tiny bit of genery. ;) )

bernhardreiter (approved):
The integration test ran fine. 
Briefly reading over the changes, there is no obvious problem.
LGTM

Issue #344, opened 2023-02-28
Shall a lockfile be used by default in the aggregator


bernhardreiter:
Current default in the aggregator is to not use a lockfile. 
It means that two instances with the same configuration can be started any may interfere with each other.
This could happen if one interim scan takes too long and the next is already started.
Is this the right default?

tschmidtb51:
> Is this the right default?

Potentially not. I'm open to that change.

bernhardreiter:
We want to change the default to using a lock file. 

* Shall be pointed out in the release notes.
* okay with 2.2.0 version number
* Can be switched off by explicit configuration.

JanHoefelmeyer:
If a lock file is configured but the option to not use a lock file is also set to true, how do we want to handle this?

 * Use configured lock file and have the switch only be effective if no lock file was given?
 * Use no lock file, neither the default one nor the explicitely configured one?
 * Throw an error because having both configured points to a faulty config?

tschmidtb51:
IMHO following the principle of least astonishment, I would expect the 2 option ("I said 'no lock file' so it is no lock file even if I told you where to put it if I would have wanted one.")

s-l-teichmann:
Having an extra flag `no_lock` to say that the otherwise configured 'lock-file` should be ignored seem a bit unfortunate and contradictory. Why not the following:

* If no explicit `lock-file` use the default.
* If `lock-file` is the empty string don't use the lock file.
* If `lock-file` is not empty use this as the path.

tschmidtb51:
@s-l-teichmann I guess my answer lead into that direction. I'm happy with both approaches. So - I would suggest that @s-l-teichmann decides which one should be implemented (as you are closer to the code and @JanHoefelmeyer).

tschmidtb51:
@JanHoefelmeyer Please consult with @s-l-teichmann and update #351 if necessary. Please also document the decision (either back in draft mode or an explicit "@tschmidtb51 ready for review")  so that I know whether I should review #351 or whether you are working on that again.

Thanks!

PR #351, opened 2023-03-20
Lockfile aggregator

JanHoefelmeyer:
Adds option no_lock to csaf_aggregator. If set to true, no lockfile will be used. If no_lock is false, the configs lockfile will be used. If no lockfile is specified, /var/csaf_aggregator/run.lock will be used as default.

Solves https://github.com/csaf-poc/csaf_distribution/issues/344

### DIFF
--- a/cmd/csaf_aggregator/config.go
+++ b/cmd/csaf_aggregator/config.go
@@ -267,8 +267,10 @@ func (c *config) setDefaults() {
 		c.Domain = defaultDomain
 	}
 
-	if !c.NoLock && c.LockFile == nil {
-		c.LockFile = &defaultLockFile
+	if c.NoLock {
+		c.LockFile = nil
+	} else if c.LockFile == nil {
+		c.LockFile = &defautLockFile
 	}
 
 	if c.Workers <= 0 {
--- a/cmd/csaf_aggregator/config.go
+++ b/cmd/csaf_aggregator/config.go
@@ -32,10 +32,7 @@ const (
 	defaultWeb            = "/var/www/html"
 	defaultDomain         = "https://example.com"
 	defaultUpdateInterval = "on best effort"
-)
-
-var (
-	defaultLockFile = "/var/csaf_aggregator/run.lock"
+	defaultLockFile       = "/var/csaf_aggregator/run.lock"
 )
 
 type provider struct {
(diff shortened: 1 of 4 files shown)


cmd/csaf_aggregator/config.go
```diff
@@ -32,10 +32,7 @@ const (
 	defaultWeb            = "/var/www/html"
 	defaultDomain         = "https://example.com"
 	defaultUpdateInterval = "on best effort"
-)
-
-var (
-	defaultLockFile = "/var/csaf_aggregator/run.lock"
+	defaultLockFile       = "/var/csaf_aggregator/run.lock"
```
tschmidtb51, line 35: I'm curious: Is there a specific reason why we choose `/var/csaf_aggregator/run.lock` for the lock file and not `/var/lock/<something>`?

tschmidtb51 (commented):
A few comments here:
1. As we change the default behavior, I guess it make sense to create the lock-file if it does not exist, correct?
2. I guess we should also improve the error message: "cannot lock to file /var/csaf_aggregator/run.lock" => "cannot acquire file lock at <PATH>. Maybe the CSAF aggregator is already running?"
3. I'm curious: Is there a specific reason why we choose /var/csaf_aggregator/run.lock for the lock file and not /var/lock/<something>?

tschmidtb51:
Just discussed with @s-l-teichmann:
1. Yes. We should create it.
2. Please implement.
3. Use `/var/lock/csaf_aggregator.lock` or /var/lock/csaf_aggregator/lock`

tschmidtb51 (changes requested):
Just discussed with @s-l-teichmann:
1. Yes. We should create it.
2. Please implement.
3. Use `/var/lock/csaf_aggregator.lock` or `/var/lock/csaf_aggregator/lock`

s-l-teichmann:
> Just discussed with @s-l-teichmann:
> 
>     1. Yes. We should create it.
> 
>     2. Please implement.
> 
>     3. Use `/var/lock/csaf_aggregator.lock` or `/var/lock/csaf_aggregator/lock`

to 1.

If I read the [sources](https://github.com/gofrs/flock/blob/6f010d1acea74a32f2f2066bfe324c08bbee30e3/flock.go#L128) of the library correctly, there is no need to create the file manually, as it
is already done by the library.

s-l-teichmann (approved):
LGTM

tschmidtb51 (approved):
LGTM and works

Issue #353, opened 2023-03-24
Add `revive` to GitHub Action


tschmidtb51:
The tool `revive` is useful to detect unused variables and other things. We should also integrate that into the GitHub Actions.

PR #355, opened 2023-03-27
Add revive action to workflows

JanHoefelmeyer:
What:

Add revive action to github actions as described in https://github.com/marketplace/actions/revive-action 

Why:

Revive can catch potential mistakes and poorly written code. Solves  https://github.com/csaf-poc/csaf_distribution/issues/353

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -30,5 +30,8 @@ jobs:
       - name: golint
         uses: Jerome1337/golint-action@v1.0.2
 
+      - name: Revive Action
+        uses: morphy2k/revive-action@v2.5.1
+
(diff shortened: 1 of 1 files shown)


s-l-teichmann (approved):
LGTM

tschmidtb51 (approved):
In general, this looks good to me; but I haven't seen the output yet (obviously, as this GH Action was not triggered since no Go file was changed). So I suggest to merge this but leave #353 open until the next release to double check whether this action need some tweaking...  (so that we see results in a way that helps us spot mistakes easy. This might also end up in a case that we need an extra action for this.)

tl;dr: merge now, check impact in/on next PRs

Issue #338, opened 2023-02-10
Avoid misunderstandings in requirement 1


tschmidtb51:
If no remote validator is given, the `csaf_checker` emits:
```
    - INFO: No remote validator configured
    - INFO: All advisories validated fine.
 ```

However, this can cause misunderstandings as the mandatory tests can't be executed without the remote validator (at least right now...).

I suggest we change the text for this case:
```
    - WARN: No remote validator configured
    - INFO: All advisories validated fine against the schema.
 ```

Even if the remote validator is configured, we need to clarify the result as one can just execute a specific set of tests (e.g. just the informative) that won't invalidate the CSAF documents.

_tested with csaf_checker v2.0.1-60-g05ad714_

PR #356, opened 2023-03-27
Rephrase csaf validation result

JanHoefelmeyer:
Give out a warning if no remote validator is configured.

Clarify that the inbuild validator only validates against the schema.

Solves https://github.com/csaf-poc/csaf_distribution/issues/338

### DIFF
--- a/cmd/csaf_checker/reporters.go
+++ b/cmd/csaf_checker/reporters.go
@@ -49,12 +49,12 @@ func (bc *baseReporter) requirement(domain *Domain) *Requirement {
 func (r *validReporter) report(p *processor, domain *Domain) {
 	req := r.requirement(domain)
 	if p.validator == nil {
-		req.message(InfoType, "No remote validator configured")
+		req.message(WarnType, "No remote validator configured")
 	}
 	if !p.invalidAdvisories.used() {
 		req.message(InfoType, "No validations performed")
 	} else if len(p.invalidAdvisories) == 0 {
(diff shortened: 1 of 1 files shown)


cmd/csaf_checker/reporters.go
```diff
@@ -49,12 +49,12 @@ func (bc *baseReporter) requirement(domain *Domain) *Requirement {
...
 	if p.validator == nil {
-		req.message(InfoType, "No remote validator configured")
+		req.message(WarnType, "No remote validator configured")
 	}
 	if !p.invalidAdvisories.used() {
 		req.message(InfoType, "No validations performed")
 	} else if len(p.invalidAdvisories) == 0 {
-		req.message(InfoType, "All advisories validated fine.")
+		req.message(InfoType, "All advisories validated fine against the schema.")
 	} else {
```
tschmidtb51, line 58: I think we need to extend the logic here:
If a remote validator and one of the presets `basic`, `mandatory`, `extended` or `full` was used, we should say `"All advisories validated fine."`
If a remote validator and none of the presets `basic`, `mandatory`, `extended` or `full` was used or no remote validator was used, we should say `"All advisories validated fine against the schema."`

tschmidtb51 (changes requested):
Please see my comment.

s-l-teichmann (approved):
LGTM

tschmidtb51 (approved):
LGTM and works

Issue #359, opened 2023-04-26
Document default  for Rate options consistently


bernhardreiter:
Code and test shows that the downloader has a default limiting rate of 1 in main.

But `--help` does not say for the downloader and the checker and the documentaton for aggregator says unlimited, while the code assumes it is limited.

The documentation shall match the code and show the default.

bernhardreiter:
I consider it part of release 2.2.0 to fix this.

s-l-teichmann:
I have looked at this issue and it is invalid.
Have a look at https://github.com/csaf-poc/csaf_distribution/blob/3590cf1ef2b0d2945ed523a63b4f6a439b7f2d8e/cmd/csaf_downloader/downloader.go#L108

The limiter is only inserted into the client chain if rate is speciefied (!= nil).

@bernhardreiter The throttling in case of redhat.com seems to be forced by the redhat server. This is not our limit! 

-> closed

PR #360, opened 2023-04-26
Update documentation about downloader rate from unlimited to 1

JanHoefelmeyer:
Solves https://github.com/csaf-poc/csaf_distribution/issues/359


s-l-teichmann (changes requested):
Defaulting to 1 means 1 request per second. This is clearly not the default we want.

bernhardreiter:
> Defaulting to 1 means 1 request per second.

But this is how it is implemented as far as I can see and the documentation should match.

> This is clearly not the default we want.

Which is a different matter, as I haven't had the time to check why this change from unlimited to 1 in the first place. Feel free to open a new issue. :)

s-l-teichmann (changes requested):
We are solving the wrong problem here.
 
I am strictly against this change.

s-l-teichmann:
This solves a problem that do not exist.

PR #363, opened 2023-04-26
Add concurrent downloads to downloader.

s-l-teichmann:
Add `--worker` flag to downloader.

Some sites seems to limit the number of requests per time interval per http client.
This PR adds a flag to download more than one advisory at once via multiple
http clients to speed up the process. By default only two clients are used.

Furthermore this PR also adds the feature to stop the downloader by `Ctrl-C`ing.

When the `ignoresigcheck` flag is set the results of the signature checks are ignored.

Bumped Go to 1.20 to use `errors.Join`.

### DIFF
--- a/cmd/csaf_downloader/downloader.go
+++ b/cmd/csaf_downloader/downloader.go
@@ -154,6 +154,55 @@ func (d *downloader) download(ctx context.Context, domain string) error {
 	})
 }
 
+func (d *downloader) downloadFiles(
+	ctx context.Context,
+	label csaf.TLPLabel,
+	files []csaf.AdvisoryFile,
+) error {
+
+	var (
+		advisoryCh = make(chan csaf.AdvisoryFile)
+		errorCh    = make(chan error)
+		errDone    = make(chan struct{})
+		errs       []error
+		wg         sync.WaitGroup
+	)
+
+	// collect errors
+	go func() {
+		defer close(errDone)
+		for err := range errorCh {
+			errs = append(errs, err)
+		}
+	}()
+
+	var n int
+	if n = d.opts.Worker; n < 1 {
+		n = runtime.NumCPU()
+	}
+
+	for i := 0; i < n; i++ {
+		wg.Add(1)
+		go d.downloadWorker(ctx, &wg, label, advisoryCh, errorCh)
+	}
+
+allFiles:
+	for _, file := range files {
+		select {
+		case advisoryCh <- file:
+		case <-ctx.Done():
+			break allFiles
+		}
+	}
+
+	close(advisoryCh)
+	wg.Wait()
+	close(errorCh)
+	<-errDone
+
+	return errors.Join(errs...)
+}
+
 func (d *downloader) loadOpenPGPKeys(
 	client util.Client,
 	doc any,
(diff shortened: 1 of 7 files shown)


docs/csaf_downloader.md
```diff
@@ -12,6 +12,7 @@ Application Options:
       --version                Display version of the binary
   -v, --verbose                Verbose output
   -r, --rate=                  The average upper limit of https operations per second
+  -w, --worker=                Number of concurrent downloads (default: 1)
```
JanHoefelmeyer, line 15: Maybe add that using not positive or too large numbers will try to use the maximum amount of workers avaible. E.g. 

    If worker is set to 0 or less, the maximum amount of available workers will be used.

Down there with the two other explanatory lines?

cmd/csaf_downloader/main.go
```diff
@@ -10,21 +10,26 @@
...
 	"log"
 	"net/http"
 	"os"
+	"os/signal"
 
 	"github.com/csaf-poc/csaf_distribution/util"
 	"github.com/jessevdk/go-flags"
 )
 
+const defaultWorker = 1
```
bernhardreiter, line 24: Shouldn't we use a default with at least 2 to demonstrate this advantage (and get more testing of it)?

cmd/csaf_downloader/downloader.go
```diff
@@ -391,56 +462,54 @@ func (d *downloader) downloadFiles(label csaf.TLPLabel, files []csaf.AdvisoryFil
...
-	return nil
+func (d *downloader) mkdirAll(path string, perm os.FileMode) error {
+	d.mkdirMu.Lock()
+	defer d.mkdirMu.Unlock()
+	return os.MkdirAll(path, perm)
 }
 
-func (d *downloader) checkSignature(data []byte, sign *crypto.PGPSignature) bool {
+func (d *downloader) checkSignature(data []byte, sign *crypto.PGPSignature) error {
+	if d.opts.Insecure {
```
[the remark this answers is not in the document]
s-l-teichmann: I totally agree. Commit 3da3b4877889e20b7d691c8792fd70e0a186db01 introduces a special flag:
```
--ignoresigcheck         Ignore signature check results, just warn on mismatch
```

docs/csaf_downloader.md
s-l-teichmann: I gave that a second thought and I have removed the CPU stuff entirely.
In commit 3da3b4877889e20b7d691c8792fd70e0a186db01 the lower bound is now 1.

s-l-teichmann:
> Shouldn't we use a default with at least 2 to demonstrate this advantage (and get more testing of it)?

We could ... I'm a bit undecided how conservative our default setting should be. @tschmidtb51 ?

JanHoefelmeyer (dismissed):
LGTM

tschmidtb51:
> Bumped Go to 1.20 to use `errors.Join`.

Please add an extra note for that to the release notes...

tschmidtb51:
> We could ... I'm a bit undecided how conservative our default setting should be. @tschmidtb51 ?

In general, I liked the idea of doing it based on the available CPUs but I understand that this might be harder to explain... IMHO, we should set the default (at least) to `2` but allow to reduce it to `1`.

s-l-teichmann:
> > Bumped Go to 1.20 to use `errors.Join`.
> 
> Please add an extra note for that to the release notes...

Will do.

s-l-teichmann:
> > We could ... I'm a bit undecided how conservative our default setting should be. @tschmidtb51 ?
> 
> In general, I liked the idea of doing it based on the available CPUs but I understand that this might be harder to explain... IMHO, we should set the default (at least) to `2` but allow to reduce it to `1`.

Commit c98fafe0a1598c864519d9ff415d594aaa9b787c sets the default to two.

tschmidtb51 (approved):
LGTM and works

Issue #325, opened 2023-01-30
Implement filename computation


tschmidtb51:
We should implement the filename computation based on the `/document/tracking/id` and use that to check whether the filename is correct. See also [section 5.1](https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#51-filename).

Currently, we just check whether the filename contains invalid characters...

tschmidtb51:
Please see [section 5.1]() for instructions on how to implement it.

IMHO, there is also an import remark:
>  The regex `[^+\-a-z0-9]+` can be used to find a character sequence which has to be replaced by an underscore. However, it SHALL NOT be applied before completing the first step.
>
> Even though the underscore _ (0x5F) is a valid character in the filename it is replaced to avoid situations where the conversion rule might lead to multiple consecutive underscores. As a result, a `/document/tracking/id` with the value 2022_#01-A is converted into 2022_01-a instead of 2022__01-a.

bernhard-herzog:
I've started to implement this in the `filename-id` branch.

PR #365, opened 2023-05-11
Check that filename matches /document/tracking/id

bernhard-herzog:
Implement #325: Check that the filename matches the ID according to the
rules in Section 5.1. All places where the code already checks the
filename syntax the code now checks that it also matches the ID.

There's one exception to this in the `csaf_uploader` where the this
additional check is only done when schema checks are enabled. That was
done for simplicity of implementation because only in that case does the
uploader parse the CSAF file.


JanHoefelmeyer (approved):
Works

Issue #251, opened 2022-07-22
Missleading error in requirement 7


mfd2007:
"version": "0.9.4-30-g0375e22",

If the provider-metadata.json is a non-valid json document, the following result is returned.

```
{
          "num": 7,
          "description": "provider-metadata.json",
          "messages": [
            {
              "type": 1,
              "text": "https://csaf-provider-1.csaf/.well-known/csaf/provider-metadata.json not found."
            },
            {
              "type": 1,
              "text": "https://csaf-provider-1.csaf/.well-known/security.txt failed to load."
            },
            {
              "type": 1,
              "text": "https://csaf.data.security.csaf-provider-1.csaf not found."
            },
            {
              "type": 2,
              "text": "No valid provider-metadata.json found."
            },
            {
              "type": 2,
              "text": "STOPPING here - cannot perform other checks."
            }
          ]
        },
```

I would expect a hint that the provider-metadata.json could not be parsed. Verbose-mode prints ```error: no document to extract data from ``` in this case.

bernhardreiter:
@s-l-teichmann did we see this as part of the rework how problems are reported with #304 #305?

JanHoefelmeyer:
Due to https://github.com/csaf-poc/csaf_distribution/pull/370 and the general checker rework, this problem can no longer occur.

Issue #305, opened 2022-11-15
Checker: Summary for provider role


Id3aFly:
Currently the checker issues at our side 3x success ("type": 0) for requirements 8, 9 and 10 with following messages:

"Performed no in-depth test of security.txt."
"Since no valid provider-metadata.json was found, no extended check was performed."
"No check about contents from https://csaf.data.security.DOMAIN performed."

Each message on its own does not look to me like a "success".

Even more, as at least one of requirements 8-10 is required for the CSAF provider, it should somewhere be mentioned, that none of these checks was successful, but at least one is required.

JanHoefelmeyer:
Should be solved by https://github.com/csaf-poc/csaf_distribution/pull/370 and the general checker rework.

PR #370, opened 2023-06-02
Fix pmd crash

JanHoefelmeyer:
Currently, trying to check an invalid provider-metadata.json leads to a crash. (It causes a nullpointerreference when ranging over *domain.Role in line 250 of csaf_checker/processor.go): 

    for _, r := range buildReporters(*domain.Role) {
        r.report(p, domain)
    }

This PR adds logging output that the provider-metadata.json of that domain couldn't be parsed and then continues onto the next domain instead of trying to build relevant reporters.

This should also address https://github.com/csaf-poc/csaf_distribution/issues/251

### DIFF
--- a/cmd/csaf_checker/processor.go
+++ b/cmd/csaf_checker/processor.go
@@ -231,13 +231,16 @@ func (p *processor) run(domains []string) (*Report, error) {
 	}
 
 	for _, d := range domains {
-		if p.checkProviderMetadata(d) {
-			if err := p.checkDomain(d); err != nil {
-				if err == errContinue || err == errStop {
-					continue
-				}
-				return nil, err
+		if !p.checkProviderMetadata(d) {
+			// We cannot build a report if the provider metadata cannot be parsed.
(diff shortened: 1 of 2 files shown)


s-l-teichmann (approved):
LGTM

JanHoefelmeyer:
Since reporters are build based off of the provider-metadata.json, which in turn is not avaible if all lookup methods fail, this PR already solves most of https://github.com/csaf-poc/csaf_distribution/issues/305. All thats left is to give out a warning instead of an error should any singular lookup method not be done, which is the content of https://github.com/csaf-poc/csaf_distribution/pull/370/commits/c3ef8e604c813bb65385b0e1dbbc7d7c85c63096. As such, this PR should also solve https://github.com/csaf-poc/csaf_distribution/issues/305

s-l-teichmann (approved):
LGTM

bernhard-herzog:
LGTM. I tested that the segmentation violation does indeed not occur anymore and the code produces a suitable log message instead

bernhard-herzog (approved):

PR #371, opened 2023-06-02
Fix go.mod and internal dependencies for major verison v2

bernhardreiter:
* Use a "/v2" in the module path to match the git version tag which lead with a 2. Change all mention of the module as dependency internally as well.
 * This is needed so that others can use this code as a dependency, if they want to.

See  https://go.dev/doc/modules/major-version


JanHoefelmeyer (approved):
LGTM

s-l-teichmann (approved):
LGTM ... BTW we currently need at least Go 1.20, but the README still talks about 1.19.1. That should be material for another PR.

Issue #375, opened 2023-06-16
`--version` broken 


tschmidtb51:
If I use `--version` on the `csaf_downloader` or `csaf_checker` (most likely also valid for other tools) it returns `0.0.0` but the `make` uses the correct one `go build -o bin-linux-amd64/  -ldflags "-X github.com/csaf-poc/csaf_distribution/util.SemVersion=2.1.1-57-g7eae607" -v ./cmd/..`

s-l-teichmann:
Will have a look at this.

s-l-teichmann:
`go build -o bin-linux-amd64/  -ldflags "-X github.com/csaf-poc/csaf_distribution/v2/util.SemVersion=2.1.1-57-g7eae607" -v ./cmd/...` works. The `v2` is missing. This is a result of PR #371

JanHoefelmeyer:
Should be solved by  https://github.com/csaf-poc/csaf_distribution/pull/378

PR #378, opened 2023-06-20
Burn v2 version into binaries.

s-l-teichmann:
Solves #375


JanHoefelmeyer (approved):
LGTM and works

tschmidtb51 (approved):
LGTM and works

Issue #221, opened 2022-07-14
Evaluate `role` to determine overall result


tschmidtb51:
The `csaf_checker` should evaluate the `role` from the `provider-metadata.json` to determine the overall result of the check.

@bernhardreiter: Please check whether that is in scope...

s-l-teichmann:
@tschmidtb51 How is this related to issue #42 ?

s-l-teichmann:
@tschmidtb51 That helps. Thanks!

bernhardreiter:
Did check the contract documents and the current CSAF specs briefly. From my point of view it is out of scope for 1.0, because it is a larger change and it is not directly written down as requirement. And the situation changed, when after feedback between cs01 und cs02 the algorithm and the recommendation how to find a valid provider-metadata.json were clarified.

We believe it is useful to make the check more dynamic depending on the role that is given in the PMD.  But it seems unwise to do it close to the 1.0.0 release or in scope for what was a proof of concept originally.

One improvement we aim for (outside of this issue) is to warn instead of pointing out as an error, if we find a valid PMD.

bernhardreiter:
Added a hint in the documentation 6a605fdbcc6b2f14220327d699ffbf77febd4e51
And we shall address #220.

PR #383, opened 2023-06-28
docs: Add hints to illustrate that the output of some skripts may be …

ThomasJunk:
…correct even though the output contains the word 'ERROR'

### DIFF
--- a/docs/scripts/Readme.md
+++ b/docs/scripts/Readme.md
@@ -23,7 +23,7 @@ Calling example (as root):
     source ./TLSConfigsForITest.sh
     set +e  # for an interactive shell, reverse set -e done by previous line
     ./TLSClientConfigsForITest.sh
-    ./setupProviderForITest.sh
+    ./setupProviderForITest.sh # calls upon the checker tool, whose output contains the word "ERROR"
     ./testAggregator.sh
     ./testDownloader.sh
 ```


docs/scripts/Readme.md
JanHoefelmeyer, line 1: I recommend being more precise. The Error in setupProviderForITest.sh should only happen during the call of [testChecker.sh](https://github.com/csaf-poc/csaf_distribution/blob/main/docs/scripts/testChecker.sh), where the minimal provider setup means the security.txt is not written. If an error happens at another point, it shouldn't be ignored. 

I propose a comment along the lines of 

    # calls upon the checker tool, whose output contains the word "ERROR"

JanHoefelmeyer: Since the error is depending on this TODO :

	for i := range lpmd.Messages {
		// TODO: Filter depending on the role.
		p.badProviderMetadata.error(lpmd.Messages[i].Message)
	}

(Currently all messages result in an error, even if they should only be warnings or infos)

The TODO should probably also be adjusted to include removal of the comment, since there should no longer be an error after this gets fixed

JanHoefelmeyer (approved):
Implementing this seems sensible until the underlying issue is fixed, which seems to be part of https://github.com/csaf-poc/csaf_distribution/issues/221

PR #388, opened 2023-06-30
Use Set type

s-l-teichmann:
As we are using generics for a while now and we recently introduced
a Set type unify all set operations to that type.


JanHoefelmeyer (approved):
LGTM

ThomasJunk (approved):
LGTM

PR #390, opened 2023-07-05
Check for advisoryLabel instead of feedlabel

JanHoefelmeyer:
Currently the check on proper access-protectedness is done via checking the TLP-level of the feed. Any advisory placed within the wrong feed would potentially be missed or misdiagnosed.  This PR causes the check to check for the TLP-level of the advisory instead.

### DIFF
--- a/cmd/csaf_checker/processor.go
+++ b/cmd/csaf_checker/processor.go
@@ -40,10 +40,11 @@ import (
 type topicMessages []Message
 
 type processor struct {
-	opts      *options
-	validator csaf.RemoteValidator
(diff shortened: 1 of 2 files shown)


ThomasJunk (dismissed):
LGTM

s-l-teichmann (changes requested):
It's more complicated. The open http client logic is affected as well.
I will re-work it.

s-l-teichmann (approved):
LGTM now.

ThomasJunk (approved):
OK!

PR #405, opened 2023-07-19
Use TOML as config file format in downloader

s-l-teichmann:
As we need more complex configuration in the downloader (and checker) with per domain options switch to TOML as config file format in the checker.


JanHoefelmeyer (approved):
LGTM & Works

ThomasJunk (approved):
LGTM & works.

Issue #402, opened 2023-07-12
Improve message in requirement 4


tschmidtb51:
We should separate the different cases in the report message:
1. We didn't had credentials and there was no access protected feed listed. (=> unlikely that a TLP:WHITE was hidden)
2. We didn't had credentials and there was at least one access protected feed listed.
3. We had credentials and didn't found an access-protected TLP:WHITE CSAF document.

JanHoefelmeyer:
After thinking about it:

About 2: I don't think we should consider feeds or advisories we were not able to access. There is already a warning and adding a "They might have been TLP:WHITE" seems redundant. They might've also had other faults we would have to report as well then. We shouldn't make assumptions about advisories we cannot access.

Similarly, there is little difference between 1. and 3.:
We did not find access-protected TLP:WHITE feeds. The user should know whether they used credentials or not. And if they used the wrong credentials, it either doesn't matter for TLP:WHITE (no access-protected feeds) or is already tracked (if the server returns a forbidden). 

As such, the current implementation seems to suffice.

PR #406, opened 2023-07-20
Track whether files could not be accessed and report it when reporting about accessibility of TLP:WHITE advisories

JanHoefelmeyer:
New processor variable "foundForbidden" tracks whether the checker was able to access all advisories. The badWhitePermissionsReporter then takes that into account when evaluating the report.


JanHoefelmeyer:
Solves https://github.com/csaf-poc/csaf_distribution/issues/402

bernhardreiter:
@JanHoefelmeyer can the accompanied branch be deleted? If so, please do.

PR #431, opened 2023-08-17
No longer set timestamp of version as part of go version in prepareUbuntuInstanceForITests

JanHoefelmeyer:
Currently, trying to set up a blank Ubuntu 20.4 Instance via prepare prepareUbuntuInstanceForITests.sh will fail since 

curl https://go.dev/VERSION\?m=text currently returns 

go1.21.0
time 2023-08-04T20:14:06Z

The timestamp setting the latest_go variable to an unusable state, causing the depending curl instance to throw a syntax error.

This PR will exclude the timestamp and allows fetching the correct go version.


s-l-teichmann (approved):
LGTM ... maybe passing `-s` to `curl` would suppress some noise ... but it's okay for now.

PR #444, opened 2023-08-25
downloader: improve code comments

bernhardreiter:


bernhardreiter:
(did not create it the way I wanted it)

PR #458, opened 2023-09-13
feat: log redirects

cintek:
added a callback to log redirects

### DIFF
--- a/cmd/csaf_downloader/downloader.go
+++ b/cmd/csaf_downloader/downloader.go
@@ -92,16 +92,24 @@ func (d *downloader) addStats(o *stats) {
 	d.stats.add(o)
 }
 
-// checkRedirect specifies the policy for handling redirects.
-func checkRedirect(req *http.Request, via []*http.Request) error {
-	slog.Debug(fmt.Sprintf("Redirecting to %s", req.URL))
+// logRedirect logs redirects of the http client.
+func logRedirect(req *http.Request, via []*http.Request) error {
+	vs := make([]string, len(via))
+	for i, v := range via {
+		vs[i] = v.URL.String()
+	}
+	slog.Debug("Redirecting",
+		"to", req.URL.String(),
+		"via", strings.Join(vs, " -> "))
 	return nil
 }
 
 func (d *downloader) httpClient() util.Client {
 
-	hClient := http.Client{
-		CheckRedirect: checkRedirect,
+	hClient := http.Client{}
+
+	if d.cfg.LogLevel.slogLevel() <= slog.LevelDebug {
+		hClient.CheckRedirect = logRedirect
 	}
 
 	var tlsConfig tls.Config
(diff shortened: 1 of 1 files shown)


cmd/csaf_downloader/downloader.go
```diff
@@ -92,9 +92,17 @@ func (d *downloader) addStats(o *stats) {
 	d.stats.add(o)
 }
 
+// checkRedirect specifies the policy for handling redirects.
+func checkRedirect(req *http.Request, via []*http.Request) error {
+	slog.Debug(fmt.Sprintf("Redirecting to %s", req.URL))
```
s-l-teichmann, line 97: Would like to to see the use of the `s` for structured in `slog`.
Something like:
```
slog.Debug("Redirecting", "to", req.URL.String())
```

cmd/csaf_downloader/downloader.go
```diff
@@ -92,9 +92,17 @@ func (d *downloader) addStats(o *stats) {
 	d.stats.add(o)
 }
 
+// checkRedirect specifies the policy for handling redirects.
+func checkRedirect(req *http.Request, via []*http.Request) error {
```
s-l-teichmann, line 96: Use `_` for unused parameters.
```
func checkRedirect(req *http.Request, _ []*http.Request) error
```

cmd/csaf_downloader/downloader.go
```diff
@@ -92,9 +92,17 @@ func (d *downloader) addStats(o *stats) {
...
+	slog.Debug(fmt.Sprintf("Redirecting to %s", req.URL))
+	return nil
+}
+
 func (d *downloader) httpClient() util.Client {
 
-	hClient := http.Client{}
+	hClient := http.Client{
+		CheckRedirect: checkRedirect,
+	}
```
s-l-teichmann, line 106: I would add the callback only if the log level is at least debug. Something like:
```
hClient := http.Client{}
if d.cfg.LogLevel.slogLevel() <= slog.LevelDebug {
    hClient.CheckRedirect = checkRedirect
}


```

s-l-teichmann (changes requested):

cmd/csaf_downloader/downloader.go
cintek: I would agree to the new name for the function if it would just be there to create the log but the purpose of the callback of the client is to check if it should follow a redirect.
We could:

- name the function `logRedirect` nonetheless
- create a new function `logRedirect` which is called from `checkRedirect` (which would be a very small function and therefore I can understand if this is not a good solution)
- stay with the current name (but maybe add a comment for the logging part)

s-l-teichmann: As soon as there is some actual checking we can name it check... For now there is only the logging.
so I would recommend to name it log... .

s-l-teichmann (approved):
LGTM

JanHoefelmeyer (approved):
LGTM && Works

PR #470, opened 2023-09-28
Downloader: unit test forwarder

s-l-teichmann:
Bring coverage of forwarder near 100%.

The remaining untested error paths belong to the fact that
we build the requests to the remote server in memory
with no limitations. In the future we may limit this but we
should do this earlier in the pipeline.


ThomasJunk (approved):
LGTM

JanHoefelmeyer (approved):
LGTM & works

PR #476, opened 2023-10-06
Add build for macOS

fjd-anh:
Added Support for macOS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,15 +12,15 @@ SHELL = /bin/bash
 BUILD = go build
 MKDIR = mkdir -p
 
-.PHONY: build build_linux build_win tag_checked_out mostlyclean
+.PHONY: build build_linux build_win build_mac_amd64 build_mac_arm64 tag_checked_out mostlyclean
 
 all:
-	@echo choose a target from: build build_linux build_win mostlyclean
+	@echo choose a target from: build build_linux build_win build_mac_amd64 build_mac_arm64 mostlyclean
 	@echo prepend \`make BUILDTAG=1\` to checkout the highest git tag before building
 	@echo or set BUILDTAG to a specific tag
 
 # Build all binaries
-build: build_linux build_win
+build: build_linux build_win build_mac_amd64 build_mac_arm64
 
 # if BUILDTAG == 1 set it to the highest git tag
 ifeq ($(strip $(BUILDTAG)),1)
@@ -29,7 +29,7 @@ endif
 
 ifdef BUILDTAG
 # add the git tag checkout to the requirements of our build targets
-build_linux build_win: tag_checked_out
+build_linux build_win build_mac_amd64 build_mac_arm64: tag_checked_out
 endif
 
 tag_checked_out:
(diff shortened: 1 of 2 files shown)


bernhardreiter:
Hi  @fjd-anh, thanks for creating a pull request for MacOSX support.

How many parts of the toolset did you test so far?
(I wonder also if manually running the scripts for testing are successful.)

Makefile
```diff
@@ -20,7 +20,7 @@ all:
 	@echo or set BUILDTAG to a specific tag
 
 # Build all binaries
-build: build_linux build_win
+build: build_linux build_win build_mac_intel build_mac_apple
```
bernhardreiter, line 23: I wonder if using `build_mac_amd64` and `build_mac_arm64`  would be slightly better because it refers to the GOARCH values. What do you think?

bernhardreiter:
So far the darwin builds will not be distributed as part of the release, is that on purpose? (I wonder does cross compilation work fine?)

Makefile
fjd-anh: Technically, they are better. But in a quick look, `arm64` and `amd64` are easily mixed up.
But if you think, they are better, that's OK for me.

fjd-anh:
> Hi @fjd-anh, thanks for creating a pull request for MacOS support.
> 
> How many parts of the toolset did you test so far? (I wonder also if manually running the scripts for testing are successful.)

I was specially interested on a mac version of the downloader and only tried this one.
The downloader and the validator work fine.
From the other tools, I only checked the `--help` output.

> So far the darwin builds will not be distributed as part of the release, is that on purpose? (I wonder does cross compilation work fine?)

No. I missed this. If you want to provide the biaries, I'll add this to the Makefile.

Makefile
bernhardreiter: As we use "amd64" for the distribution packages in a release, I think it is slightly better to use them here as well. But as always I think a lot about the right consistent naming... ;)

bernhardreiter:
Yes, it would be nice to have this as binary so we get it into the  release assets. Maybe just like the windows zip file, leaving out the csaf_provider and the csaf_aggregator.   Would it be okay to do one archives for both sets of MacOS' binaries? 

We need to do a test on MacOS as well and we need to ask the BSI who is funding the development of these tools, before we can integrate it. This may take a few days.

fjd-anh:
@bernhardreiter I changed the build targets and added them to the dist target.

bernhardreiter:
Our test on MacOS for the four client application will be done by @cintek

bernhardreiter:
We got the permission to activate a build and add client binaries to the release, noting that they are "community supported".

bernhardreiter:
Added a note for the readme and an archive to be added, if we do a release. So waiting for our test from @cintek  now. :)

JanHoefelmeyer (approved):
LGTM

cintek:
What I tested so far are downloader and validator and they are working as expected. Next step is to prepare the Mac to test provider and uploader.

bernhardreiter:
> What I tested so far are downloader and validator and they are working as
> expected. Next step is to prepare the Mac to test provider and uploader.

Would be find to test the validator and checker instead of the provider?
You can use a validation server and a provider on a GNU/Linux VM.

cintek:
> What I tested so far are downloader and validator and they are working as expected. Next step is to prepare the Mac to test provider and uploader.
> Would be find to test the validator and checker instead of the provider? You can use a validation server and a provider on a GNU/Linux VM.

I tested the validator already. Now I also tested the checker by using verbose and version flag, writing the output to a file and using a config file. All of this just worked fine.

bernhardreiter:
@cintek cool!

Did you also check the uploader? (To e.g. upload to a provider running on a GNU/Linux system)?

bernhardreiter (approved):
LGTM (and @cintek tested that binaries work)

JanHoefelmeyer (approved):
LGTM

PR #481, opened 2023-10-17
fix: improve logging for downloader and aggregator

bernhardreiter:
* use full name for printing out the used logfile for the downloader.
 * for debug or verbose, log the timeintervall that will be used for downloader and aggregator. (The checker has this as part of its output already.)


s-l-teichmann (approved):
LGTM

JanHoefelmeyer (approved):
LGTM & works

Issue #486, opened 2023-10-30
Relationships not evaluated


tschmidtb51:
The current version (might be v3.0.0-beta-39-g0905824) does not evaluate purls given in a `relationship` item.

s-l-teichmann:
@tschmidtb51 Can you describe briefly what the expected behavior should be?

tschmidtb51:
All purls in the following example should be found:
```
{
  "document": {
    "title": "CSAF_searcher purl test",
    "csaf_version": "2.0",
    "category": "csaf_base",
    "publisher": {
      "category": "vendor",
      "name": "Example Company",
      "namespace": "https://www.example.com",
      "contact_details": "psirt@example.com"
    },
    "tracking": {
      "revision_history": [
        {
          "number": "1",
          "date": "2023-11-02T00:00:00Z",
          "summary": "Initial version"
        }
      ],
      "initial_release_date": "2023-10-30T00:00:00Z",
      "current_release_date": "2023-10-30T00:00:00Z",
      "version": "1",
      "id": "ESA-2023-1003",
      "status": "draft"
    }
  },
  "product_tree": {
    "branches": [
      {
        "category": "vendor",
        "name": "Example Company",
        "branches": [
          {
            "category": "product_name",
            "name": "Product A",
            "branches": [
              {
                "category": "product_version",
                "name": "1.0",
                "product": {
                  "name": "Product A 1.0",
                  "product_id": "1",
                  "product_identification_helper": {
                    "purl": "pkg:example/a:1.0"
                  }
                }
              }
            ]
          }
        ]
      }
    ],
    "full_product_names": [
      {
        "name": "Any",
        "product_id": "CSAFPID-0002",
        "product_identification_helper": {
          "purl": "pkg:no-valid-type/any"
        }
      },
      {
        "name": "Other",
        "product_id": "CSAFPID-0003",
        "product_identification_helper": {
          "purl": "pkg:no-valid-type/other"
        }
      }
    ],
    "relationships": [
      {
        "category": "installed_with",
        "product_reference": "CSAFPID-0002",
        "relates_to_product_reference": "CSAFPID-0003",
        "full_product_name": {
          "name": "Any installed with Other",
          "product_id": "CSAFPID-0004",
          "product_identification_helper": {
            "purl": "pkg:no-valid-type/any-other"
          }
        }
      }
    ]
  }
}
```
Currently, `pkg:no-valid-type/any-other` is missing.

[...]

PR #497, opened 2023-11-02
API: Fix pattern matching of purls and document categories in advisory model

s-l-teichmann:
Extract purls from relationships in purls_searcher example.

Solves #486


tschmidtb51 (approved):
LGTM and works

JanHoefelmeyer (approved):
LGTM

Issue #474, opened 2023-09-29
Schema validation seems to ignore formats


tschmidtb51:
It looks like the current version (3.0.0-beta1; but it might have existed in previous version), does not evaluated the format contraints. This was discovered when checking https://github.com/cisagov/CSAF/blob/971b6976d1ce52a710025aef920dba72d3a0675e/csaf_files/OT/white/2023/icsa-23-159-02.json#L269 resulted in a different amoung of errors while using the `csaf_validator` with the `csaf_validator-service` preset of `mandatory` (0) vs `basic` (1).

We need to find out, why the `uri` wasn't checked for the format.

s-l-teichmann:
Turns out that the [Schema Compiler](https://pkg.go.dev/github.com/santhosh-tekuri/jsonschema/v5#Compiler) from the validation library we use has a flag to check the formats.
It is off by default.

PR #475 turns this flag on.

**Without** the PR:
```
./csaf_validator icsa-23-159-02.json
"icsa-23-159-02.json" passes the schema validation.
```

**With** the PR:
```
./csaf_validator icsa-23-159-02.json
schema validation errors of "icsa-23-159-02.json"
  * https://docs.oasis-open.org/csaf/csaf/v2.0/csaf_json_schema.json#: doesn't validate with https://docs.oasis-open.org/csaf/csaf/v2.0/csaf_json_schema.json#
  * /vulnerabilities/0/remediations/3/url: 'www.illustracameras.com' is not valid 'uri'
```

I'm not totally sure what the overall consequences of always turning on this flag are but I think we should go with it.

tschmidtb51:
This seem not to be resolved for the following file:
```
{
  "document": {
    "category": "csaf_base",
    "csaf_version": "2.0",
    "publisher": {
      "category": "other",
      "name": "Test",
      "namespace": "https://www.example.com"
    },
    "references": [
      {
        "summary": "Invalid URL",
        "url": "https://security.business.xerox.com/wp-content/uploads/2022/11/Xerox-Security-Bulletin-XRX22-026-FreeFlow®-Print-Server-v7.pdf"
      }
    ],
    "title": "Testfile with invalid URI",
    "tracking": {
      "current_release_date": "2023-11-30T12:25:34.622Z",
      "generator": {
        "date": "2023-11-30T12:27:08.481Z",
        "engine": {
          "name": "Secvisogram",
          "version": ".2.2.15"
        }
      },
      "id": "Test-2023-11-30",
[...]

PR #517, opened 2023-12-01
Validating URIs in JSON Schema like validator service.

s-l-teichmann:
The validator service validates the 'uri' format with a regular expression and does not allow encoding other than ASCII in this strings.
This PR emulates this behavior .

Solves #474

### DIFF
--- a/csaf/validation.go
+++ b/csaf/validation.go
@@ -12,9 +12,11 @@ import (
 	"bytes"
 	_ "embed" // Used for embedding.
 	"io"
+	"regexp"
 	"sort"
 	"strings"
 	"sync"
+	"unicode"
 
 	"github.com/santhosh-tekuri/jsonschema/v5"
 )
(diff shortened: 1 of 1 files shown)


tschmidtb51:
@bernhardreiter You mentioned in https://github.com/csaf-poc/csaf_distribution/issues/474#issuecomment-1835729531 to split the issue and address the problem upstream. Should we go ahead with this PR and then do the upstream thing or should we just track it here and hope for upstream to resolve the issue?

s-l-teichmann:
> @bernhardreiter You mentioned in [#474 (comment)](https://github.com/csaf-poc/csaf_distribution/issues/474#issuecomment-1835729531) to split the issue and address the problem upstream. Should we go ahead with this PR and then do the upstream thing or should we just track it here and hope for upstream to resolve the issue?

I will do the split out tomorrow.

bernhardreiter:
@s-l-teichmann did you do the split-up?

bernhardreiter:
Considering the implementation options:
as `format.validateURI` is not a public method and it is used in a number of places, it is much better to fix this upstream.

So https://github.com/santhosh-tekuri/jsonschema/pull/226 tries to do that

We shall put more efforts then into into migrating upstream from v5 to v6.

PR #523, opened 2024-01-08
Checker: Ensure that the processor is reset before checking each domain.

### LINKED ISSUE #522: CSAF checker: mixing domains, failing validations
During the CSAF workshop one task was to use `csaf_checker` to download and validate CSAF documents. As we had multiple (more than 10) "domains", I was using the command like `csaf_checker dns1 dns2 dns3 …`.

That resulted in successful downloads, but failing validations for some domains, due to DNS in mismatches. Validating all domains with one domain each worked though.

It somehow feels like this is suffering from: https://go.dev/doc/faq#closures_and_goroutines … but maybe it is something different.

s-l-teichmann:
In issue #522 there are signs that the checker get mixed up when checking multiple domains.

One possible reason could be the not proper reset of the processor before checking
a particular. This PR guarantees that the reset is done before the check.

Maybe fixes #522

### DIFF
--- a/cmd/csaf_checker/processor.go
+++ b/cmd/csaf_checker/processor.go
@@ -202,8 +202,8 @@ func (p *processor) close() {
 	}
 }
 
-// clean clears the fields values of the given processor.
-func (p *processor) clean() {
+// reset clears the fields values of the given processor.
+func (p *processor) reset() {
 	p.redirects = nil
 	p.noneTLS = nil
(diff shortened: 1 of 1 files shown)


ThomasJunk (approved):
LGTM

JanHoefelmeyer (approved):
As far as I can tell, LGTM && works

PR #530, opened 2024-04-18
Added support for structured logging in `csaf_aggregator`

### LINKED ISSUE #462: Consider structured logging for aggregator, too.
The downloader now sports structure logging as its focus shifted towards an automation tool.
The aggregator always was thought as such a tool.
Therefore  the aggregator should use the structured logging, too.

oxisto:
This PR adds structured logging for the aggregator service. Currently, only the text handler is used, but I can extend this to use the JSON handler as well. In this case, probably some code that is shared between the aggregator and the downloader would need to be moved to a common package.

I was also wondering, whether this repo is moving to Go 1.21 at the future, since `slog` was introduced in to the standard lib in 1.21. So currently, this still relies on the `x/exp` package.

Fixes #462


s-l-teichmann:
Thank for your contribution. I will review the PR as soon as I've git some time.

> I was also wondering, whether this repo is moving to Go 1.21 at the future, since `slog` was introduced in to the standard lib in 1.21. So currently, this still relies on the `x/exp` package.

We clearly can do this as we already where there. The 'x/exp' stuff was only introduced in PR #514 to help backwards compatibility.

oxisto:
> Thank for your contribution. I will review the PR as soon as I've git some time.

Looking forward to the feedback!

> 
> > I was also wondering, whether this repo is moving to Go 1.21 at the future, since `slog` was introduced in to the standard lib in 1.21. So currently, this still relies on the `x/exp` package.
> 
> We clearly can do this as we already where there. The 'x/exp' stuff was only introduced in PR #514 to help backwards compatibility.

Ah I see. Well it should be more or less a case of just replacing the exp package with the `slog` package, once you decide to revisit Go 1.21.

tschmidtb51:
Just a reminder: When we update the Go version, this also needs to be documented in https://github.com/csaf-poc/csaf_distribution/blob/main/README.md#build-from-sources and https://github.com/csaf-poc/csaf_distribution/blob/main/docs/Development.md#supported-go-versions as well.

bernhardreiter:
@oxisto are you okay with the new license `Apache-2.0` for your contribution #535 ?

Can you also do the Go Version update has part of this PR?

oxisto:
> @oxisto are you okay with the new license `Apache-2.0` for your contribution #535 ?

Yes, that is ok for me.

> 
> Can you also do the Go Version update has part of this PR?

Sure, I can do that and already use `slog` instead of the x/exp one.

bernhard-herzog (commented):
I tested this briefly and it seems to work. I didn't check all the changed log statements, though.

bernhardreiter:
@oxisto a side question about Github: Is it possible for your to run the "Integration Test" workflow (https://github.com/oxisto/csaf_distribution/blob/main/.github/workflows/itest.yml ) because obviously I cannot do this on your clone and as it is not part of our automatic test I have not figured out how to run it on a pull request manually).

bernhardreiter:
@oxisto in one commit message you are asking a few questions (which ideally shouldn't be in a commit message):

> but I can extend this to use the JSON handler as well. 

It should be similar to how the downloader handels it.

> In this case, probably some code that is shared between the aggregator 
> and the downloader would need to be moved to a common package.

Yes common code should go into a common package, we already have util.

Both can be done in different PRs.

cmd/csaf_aggregator/mirror.go
```diff
@@ -359,7 +359,7 @@ func (w *worker) doMirrorTransaction() error {
 		}
 	}
 
-	log.Printf("sym link: %s -> %s\n", w.dir, target)
+	w.log.Debug("Creating symbol", "from", w.dir, "to", target)
```
bernhardreiter, line 362: should be "sym link" instead of "symbol"

bernhardreiter (changes requested):
LGTM and works.
A symbolic link is called `symbol` twice, I recommend to change this.

Removing the `verbose` command line option and adding setting for the log-level would be the next step, however it might also affect the command line, so it can be delayed.

oxisto:
> @oxisto in one commit message you are asking a few questions (which ideally shouldn't be in a commit message):
> 
> > but I can extend this to use the JSON handler as well.
> 
> It should be similar to how the downloader handels it.
> 
> > In this case, probably some code that is shared between the aggregator
> > and the downloader would need to be moved to a common package.
> 
> Yes common code should go into a common package, we already have util.
> 
> Both can be done in different PRs.

Sure, once this is merged I can have a look at what can be extracted into a common package, so that other tools can potentially use this as well and that behaviour is consistent across `cmd`.

Apart from that I have changed the symbol to sym link.

oxisto:
> @oxisto a side question about Github: Is it possible for your to run the "Integration Test" workflow (https://github.com/oxisto/csaf_distribution/blob/main/.github/workflows/itest.yml ) because obviously I cannot do this on your clone and as it is not part of our automatic test I have not figured out how to run it on a pull request manually).

https://github.com/oxisto/csaf_distribution/actions/runs/8821657205/job/24218013269

bernhardreiter (approved):
works and LGTM

bernhard-herzog (approved):
Looks good to me now

Issue #520, opened 2023-12-14
Print provider-metadata.json files per domain


bernhardreiter:
A domain can have several provider-metadata.json (PMD) files (which each in turn can have several directory and ROLIE feeds).

To expose the discovery part of downloading, to be used by other components in a scripting environment, it maybe useful to have an option just to find and output all PMDs.

Use case:
A script author wants to find if a domain offers CSAF 2.0 docs and all their PMDs to process them with no-go code/scripts.

immqu:
Hi, I would like to take care of this issue!

Here is what I understand: A provider can host several PMDs under one domain, using the three options well-known path, security.txt, and DNS-URL. Currently, the [load](https://github.com/csaf-poc/csaf_distribution/blob/main/csaf/providermetaloader.go#L114) function only looks for the first valid PMD it finds, but aggregators or other users may want to check for all PMDs that are hosted under a domain. 

So I would add a function to [providermetaloader.go](https://github.com/csaf-poc/csaf_distribution/blob/main/csaf/providermetaloader.go), which already includes most of the necessary logic to enumerate all PMD files. I'd add a new function `enumeratePMDs` which obtains the PMDs from the well-known path and the security.txt. If these two do not include any results, it should check for the DNS-URL (as defined in the [spec](https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#731-finding-provider-metadatajson)).

Additionally, the [checker](https://github.com/csaf-poc/csaf_distribution/blob/main/cmd/csaf_downloader/downloader.go) could be extended to use the new function.

bernhardreiter:
To clarify: The only way to specify one then on PMD is via the security.txt. See 
* https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#718-requirement-8-securitytxt 
* https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#731-finding-provider-metadatajson

If there is a DNS entry, it has to be the same as the well-known file. 

Several PMDs in the security.txt shall only be temporary.

oxisto:
> To clarify: The only way to specify one then on PMD is via the security.txt. See
> 
> * https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#718-requirement-8-securitytxt
[...]

PR #536, opened 2024-04-23
Print provider-metadata.json files per domain

immqu:
This PR addresses #520. It adds an enumerate method to the providermetaloader.go which loads the PMDs from the well-known path, the security.txt, or the DNS URL (as defined in the spec in [requirement 7.3.1](https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#731-finding-provider-metadatajson)).

The PR also includes usage of the new method in the csaf_downloader and its main method.

There is, however, still a doubt that I have about the issue: Why does the load method compare the PMD from the well-known path to the first one found in the security.txt? I couldn't find a respective requirement in the spec. It could be the case that a provider provides the PMD in the well-known path and only advertises additional PMDs in the security.txt, right?

Looking forward to your feedback!

### DIFF
--- a/cmd/csaf_downloader/config.go
+++ b/cmd/csaf_downloader/config.go
@@ -58,7 +58,7 @@ type config struct {
 	IgnorePattern        []string          `long:"ignore_pattern" short:"i" description:"Do not download files if their URLs match any of the given PATTERNs" value-name:"PATTERN" toml:"ignore_pattern"`
 	ExtraHeader          http.Header       `long:"header" short:"H" description:"One or more extra HTTP header fields" toml:"header"`
 
-	EnumeratePMDOnly bool `long:"enumerate_pmd_only" description:"If this flag is set to true, the donwloader will only enumerate valid provider metadata files, but not download documents" toml:"enumerate_pmd_only"`
+	EnumeratePMDOnly bool `long:"enumerate_pmd_only" description:"If this flag is set to true, the downloader will only enumerate valid provider metadata files, but not download documents" toml:"enumerate_pmd_only"`
 
 	RemoteValidator        string   `long:"validator" description:"URL to validate documents remotely" value-name:"URL" toml:"validator"`
 	RemoteValidatorCache   string   `long:"validator_cache" description:"FILE to cache remote validations" value-name:"FILE" toml:"validator_cache"`
--- a/cmd/csaf_downloader/config.go
+++ b/cmd/csaf_downloader/config.go
@@ -58,6 +58,8 @@ type config struct {
 	IgnorePattern        []string          `long:"ignore_pattern" short:"i" description:"Do not download files if their URLs match any of the given PATTERNs" value-name:"PATTERN" toml:"ignore_pattern"`
 	ExtraHeader          http.Header       `long:"header" short:"H" description:"One or more extra HTTP header fields" toml:"header"`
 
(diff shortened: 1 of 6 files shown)


cmd/csaf_downloader/config.go
```diff
@@ -58,6 +58,8 @@ type config struct {
 	IgnorePattern        []string          `long:"ignore_pattern" short:"i" description:"Do not download files if their URLs match any of the given PATTERNs" value-name:"PATTERN" toml:"ignore_pattern"`
 	ExtraHeader          http.Header       `long:"header" short:"H" description:"One or more extra HTTP header fields" toml:"header"`
 
+	EnumeratePMDOnly bool `long:"enumerate_pmd_only" description:"If this flag is set to true, the donwloader will only enumerate valid provider metadata files, but not download documents" toml:"enumerate_pmd_only"`
```
bernhardreiter, line 61: Typo "donwloader"

bernhardreiter:
Hi @immqu, thanks for your contribution!

We will review it in the next days. I've started the automatic checks.
Note that I found a typo and the linter also found something (see check report).

immqu:
Hi, I have fixed the typo and linting errors.

Also, I realized the doubt I described above about the Load method came from overlooking one sentence in the spec ("If one of the URLs fulfills requirement 9, this MUST be used as the first CSAF entry in the security.txt"). Currently, the new Enumerate method lists all the PMDs including the duplicated one from the security.txt, but I think we could leave it like that, since this allows to verify that the PMDs are provided correctly.

bernhardreiter:
@immqu thanks! Can you run the Integration Test action on your branch for an additional check? (Go to Actions and then run it manually on your branch.)

csaf/providermetaloader.go
```diff
@@ -108,8 +109,51 @@ func NewProviderMetadataLoader(client util.Client) *ProviderMetadataLoader {
 	}
 }
 
-// Load loads a provider metadata for a given path.
-// If the domain starts with `https://` it only attemps to load
```
bernhardreiter, line 112: I believe the comment for documentation "If the domain starts with `https://` it only attemps to load this as PMD" or so, should stay in.

immqu:
I have updated the fork with the new changes and successfully run the integration tests!

bernhardreiter:
docs/csaf_downloader.md needs an update for the new flag (it is done manually after running --help)

bernhardreiter (approved):
works and LGTM

(I did check the autodoc updates, though I wonder why this has not done with our commits in the main repository, but it should be an improvement if it does in this way.)

There is improvement potential for subsequent changes, like removing duplicates or changing the name of the function to something that removes the "numbering" part and potentially to merge the location functions to use one implementation of the search strategy.

csaf/providermetaloader.go
```diff
@@ -108,8 +109,51 @@ func NewProviderMetadataLoader(client util.Client) *ProviderMetadataLoader {
...
+	secResults := pmdl.loadFromSecurity(domain)
+	slog.Info("Found provider metadata results in security.txt", "num", len(secResults))
+
+	for _, result := range secResults {
+		if result.Valid() {
+			resPMDs = append(resPMDs, result)
+		}
+	}
+
+	// According to the spec, only if no PMDs have been found, the should DNS URL be used
```
ThomasJunk, line 146: This looks like a typo in comment.

Perhaps it should read

>[...]then the DNS URL should be used

or

>[...] the DNS URL should be used

ThomasJunk (approved):
LGTM.

Typo in comment should not block merging.

PR #543, opened 2024-05-07
[README.md] Enhance linking and grammar

fvsamson:
- Fix grammar and enhance wording.
- Consecutive web-links without non-linked, non-whitespace characters are … problematic; thus rectified.
- References to outdated specifications do not make much sense, hence omitted, especially as the link to the current specification immediately followed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 # csaf_distribution
 
 Implements a [CSAF](https://csaf.io/)
-([Specification v2.0](https://docs.oasis-open.org/csaf/csaf/v2.0/csaf-v2.0.html))
+([Specification v2.0](https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html) and its [errata](https://docs.oasis-open.org/csaf/csaf/v2.0/csaf-v2.0.html))
 trusted provider, checker, aggregator and downloader.
 Includes an uploader command line tool for the trusted provider.
 
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 
 # csaf_distribution
 
-Implementations of a [CSAF](https://csaf.io/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,9 +10,8 @@
 
 # csaf_distribution
 
-An implementation of a [CSAF](https://csaf.io/)
-[2.0 Spec](https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html)
-([Errata](https://docs.oasis-open.org/csaf/csaf/v2.0/csaf-v2.0.html))
+Implements a [CSAF](https://csaf.io/)
+([Specification v2.0](https://docs.oasis-open.org/csaf/csaf/v2.0/csaf-v2.0.html))
```
bernhardreiter, line 14: Hi @fvsamson, nice to read from you! :)

The problem here is that https://docs.oasis-open.org/csaf/csaf/v2.0/csaf-v2.0.html only links to the errata and there is no easily found link to the current main content of the 2.0 speficiation. It is there, but under appendix A of the errata. Also the new link to the main specification https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html, which you have deleted in this change, does not link to the errata. So your change makes it much harder to access both. 

Any idea to resolve this for good? What about making a new section that explains the standard and the link both with more explanation?

bernhardreiter (changes requested):
The "An implementation of a" -> "Implements a" is an improvement, but now a link to the main body of the standard is missing. (See comment).

README.md
fvsamson: Fixed IMO.  Please re-review.

tschmidtb51 (approved):
LGTM

bernhardreiter (approved):
Thx

PR #547, opened 2024-06-21
Remove data races in downloader caused by shared use of json path eval.

s-l-teichmann:
`util.PathEval` objects are intended to be used single threaded.
Therefore sharing one for all workers in the downloader is causing data races.

This PR addresses this.


koplas (approved):
LGTM

ThomasJunk:
LGTM

ThomasJunk (approved):
Works!

PR #531, opened 2024-04-19
Fix missing error messages from loading provider-metadata.json

mgoetzegb:
## What

Fix: don't drop error messages from loading provider-metadata.json

Additionally removed the duplicate check of provider metadata candidates retrieved from `security.txt`.

## Why

Previously in case of trying last resort dns, all other error messages were dropped.

PR #548, opened 2024-07-16
fix error message in csaf downloader

mgoetzegb:
## What

Minor fix of the error message returned on an invalid provider metadata json in the csaf downloader.

## Why

The error message had a trailing `:` which suggest that there are some details which were truncated. However the details are already printed before in the log.

The extra `:` was accidentally added in #531


koplas (approved):

JanHoefelmeyer (approved):
LGTM & works

PR #549, opened 2024-07-17
Make downloader usable as library

koplas:
This pull request refactors the CLI downloader to allow usage of it as a library.


s-l-teichmann:
Talk to @koplas : We see many conflicts here and the real task would be to create something like the continuous downloader in ISDuBA.

PR #552, opened 2024-07-24
feat: no longer require to be root user to call setup scripts

JanHoefelmeyer:
Currently you need to be root to use the setupscripts as described. This will allow users with sudo privileges to use the setup scripts as well.

### DIFF
--- a/docs/scripts/Readme.md
+++ b/docs/scripts/Readme.md
@@ -12,10 +12,10 @@ and configures nginx for serving TLS connections.
 
 As creating the folders needs to authenticate with the csaf_provider, the configurations of TLS server and Client certificate authentication should be set. So it is recommended to call the scripts in this order: `TLSConfigsForITest.sh`, `TLSClientConfigsForITest.sh`, `setupProviderForITest.sh`
 
-Calling example (as root):
+Calling example (as user with sudo privileges):
 ``` bash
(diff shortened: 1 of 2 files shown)


bernhardreiter (approved):
LGTM

koplas (approved):

docs/scripts/Readme.md
koplas, line 1: Running `source ./TLSConfigsForITest.sh` will result in user logout. Multiple runs of `TLSConfigsForITest.sh` will result in a broken nginx configuration.

koplas (changes requested):

bernhardreiter:
> Multiple runs of TLSConfigsForITest.sh will result in a broken nginx configuration.

That is known and accepted for now, as this are just helpers. So it is not subject to improvement in this merge request. You could add a hint to the readme though.

cintek (commented):
LGTM & works

cintek (approved):

docs/scripts/Readme.md
koplas: If a single command in the script fails, it will logout the user. This makes debugging more difficult.

bernhardreiter: @koplas it is a known situation that if a single command fails with "source" and this bash mode, but better than to not fail.

koplas (approved):
I still prefer a failsafe, but this can be handled in another pull request.

PR #553, opened 2024-07-25
Use a default user agent

### LINKED ISSUE #500: Create unique user-agent
Currently, we use the standard go-http-lib user agent. However, to be able to better track the usage of the tool, we should use our own user-agent string, e.g., `"<program> <version>"`.

koplas:
Closes #500

### DIFF
--- a/util/client.go
+++ b/util/client.go
@@ -61,6 +61,11 @@ func (hc *HeaderClient) Do(req *http.Request) (*http.Response, error) {
 			req.Header.Add(key, v)
 		}
 	}
+
+	// Use default user agent if none is set
+	if userAgent := hc.Header.Get("User-Agent"); userAgent == "" {
+		req.Header.Add("User-Agent", "csaf-distribution/"+SemVersion)
+	}
 	return hc.Client.Do(req)
 }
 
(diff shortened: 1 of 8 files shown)


util/client.go
```diff
@@ -61,6 +61,11 @@ func (hc *HeaderClient) Do(req *http.Request) (*http.Response, error) {
 			req.Header.Add(key, v)
 		}
 	}
+
+	// Use default user agent if none is set
+	if userAgent := hc.Header.Get("User-Agent"); userAgent == "" {
+		req.Header.Add("User-Agent", "csaf-distribution "+SemVersion)
```
bernhardreiter, line 67: The generated user agent must follow https://datatracker.ietf.org/doc/html/rfc9110#name-user-agent which it currently does not.

bernhardreiter (changes requested):
* Documentation changes are missing.
* The chosen user-agent does not seem standard compliant (see comment).

JanHoefelmeyer (approved):
LGTM

bernhardreiter (approved):
LGTM
@JanHoefelmeyer can you also do a functional test, please.

JanHoefelmeyer:
Have done a functional test, so LGTM & works

JanHoefelmeyer (approved):

Issue #555, opened 2024-08-06
Missing fingerprint in PMD leads to an error


tschmidtb51:
Today, I came across a weird report regarding OpenPGP stating:
```
Requirement 20: Public OpenPGP Key (failed)
    - ERROR: Fingerprint of public OpenPGP key https://example.com/.well-known/csaf/openpgp/4C071FE3584D1E7CD70FE8B2CF9436A769B3A45B.asc does not match remotely loaded.
    - INFO: No OpenPGP keys loaded.
```
In the same report, it says:
```
Requirement 19: Signatures
    - INFO: All signatures verified.
```

That does not make sense to me: Either the OpenPGP key couldn't be loaded (then one can't check signatures) or the OpenPGP key was loaded. Also, it is not quite clear to me what "Fingerprint of public OpenPGP key https://example.com/.well-known/csaf/openpgp/4C071FE3584D1E7CD70FE8B2CF9436A769B3A45B.asc does not match remotely loaded." mean. Where does the fingerprint come from?

*(Sorry, I needed to redact some values for privacy concerns.)*

bernhardreiter:
The message comes from our source code, here are the relevant places

https://github.com/csaf-poc/csaf_distribution/blob/8feddc70e1c945e2cf2ec8cab92525aa8e89106d/cmd/csaf_checker/processor.go#L1521-L1524

https://github.com/csaf-poc/csaf_distribution/blob/8feddc70e1c945e2cf2ec8cab92525aa8e89106d/cmd/csaf_downloader/downloader.go#L369-L374

bernhardreiter:
The code compares that for a PMD section like this:

```json
{
 "public_openpgp_keys": [
    {
      "fingerprint": "804FED63730227FF2FB6D9712EA2477380F3EDCB",
      "url": "https://intevation.de/.well-known/csaf/openpgp/804FED63730227FF2FB6D9712EA2477380F3EDCB.asc"
    }
  ]
}
```

the fingerprint calculated from the downloaded file from the URL matches the given string for `"fingerprint":`.  The matching rules are "a general form of case-insensitivity" (https://pkg.go.dev/strings#EqualFold). A widely used implementation like GnuPG accepts more fingerprint formats. So this is a possible cause for the missmatch. A further analysis should do the comparison manually.

bernhardreiter:
The following patch (untested) would give more details when run:
```diff
[...]

PR #558, opened 2024-08-08
Differentiate between empty or no fingerprint

koplas:
This pull request introduces an API change, which allows for a better diagnosis of fingerprint errors.

### DIFF
--- a/cmd/csaf_checker/processor.go
+++ b/cmd/csaf_checker/processor.go
@@ -83,10 +83,8 @@ type reporter interface {
 	report(*processor, *Domain)
 }
 
-var (
-	// errContinue indicates that the current check should continue.
-	errContinue = errors.New("continue")
-)
+// errContinue indicates that the current check should continue.
+var errContinue = errors.New("continue")
 
 type whereType byte
 
@@ -970,8 +968,7 @@ func (p *processor) checkChanges(base string, mask whereType) error {
 				continue
 			}
 			path := r[pathColumn]
-			times, files =
-				append(times, t),
+			times, files = append(times, t),
 				append(files, csaf.PlainAdvisoryFile(path))
 		}
 		return times, files, nil
@@ -1227,7 +1224,6 @@ func (p *processor) checkWhitePermissions(string) error {
 // According to the result, the respective error messages added to
 // badProviderMetadata.
 func (p *processor) checkProviderMetadata(domain string) bool {
-
 	p.badProviderMetadata.use()
 
 	client := p.httpClient()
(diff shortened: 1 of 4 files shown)


cmd/csaf_checker/processor.go
```diff
@@ -1518,8 +1518,13 @@ func (p *processor) checkPGPKeys(_ string) error {
...
 		}
 
+		if key.Fingerprint == "" {
+			p.badPGPs.warn("No fingerprint for public OpenPGP key found.")
+			continue
+		}
+
 		if !strings.EqualFold(ckey.GetFingerprint(), string(key.Fingerprint)) {
-			p.badPGPs.error("Fingerprint of public OpenPGP key %s does not match remotely loaded.", u)
+			p.badPGPs.error("Given Fingerprint ('%q') of public OpenPGP key %q does not match remotely loaded ('%q').", string(key.Fingerprint), u, ckey.GetFingerprint())
```
s-l-teichmann, line 1527: `s/'%q'/%q/g`

s-l-teichmann (changes requested):

tschmidtb51:
Is the case "no fingerprint is given for this key" differentiated from "an empty fingerprint is given for this key"?

koplas:
> Is the case "no fingerprint is given for this key" differentiated from "an empty fingerprint is given for this key"?

No, in this implementation, both cases are handled the same. To separate both cases, `csaf.PGPKey` needs to be changed, which affects the public API.

tschmidtb51:
> > Is the case "no fingerprint is given for this key" differentiated from "an empty fingerprint is given for this key"?
> 
> No, in this implementation, both cases are handled the same. To separate both cases, `csaf.PGPKey` needs to be changed, which affects the public API.

In future versions, we should differentiate that. 
@bernhardreiter / @s-l-teichmann: Please decide when it makes most sense to implemented this.

bernhardreiter:
> > Is the case "no fingerprint is given for this key" differentiated from "an empty fingerprint is given for this key"?

> To separate both cases, `csaf.PGPKey` needs to be changed, which affects the public API.

Then we should not implement it for 3.x as we would need to go to version 4 if we change the public API. @koplas can you implement a version that just gives more details for 3.x? (Please do not put in unrelated whitespace changes.)

@tschmidtb51 wrote:
> In future versions, we should differentiate that.

If I understand the schema correctly there is a ` "minLength": 40` in place. So I'd expect an empty fingerprint value would violate the schema. So I expect both cases to result into to different messages. @koplas can you test what happens with an empty fingerprint value in the current code?

bernhardreiter:
(@s-l-teichmann sorry for the back and forth of your review. It has been addressed and I am trying to remove you as a requested reviewer, which somehow does not work.)

koplas:
> If I understand the schema correctly there is a ` "minLength": 40` in place. So I'd expect an empty fingerprint value would violate the schema. So I expect both cases to result into to different messages. @koplas can you test what happens with an empty fingerprint value in the current code?

If the fingerprint is empty, the `csaf_checker` prints the following error message: "Could not parse the Provider-Metadata.json".
Using the `csaf_downloader` will result in "no valid provider-metadata.json found".
No additional error messages are printed with this branch and the schema validation prevents the check for empty fingerprints.

bernhardreiter:
As discussed in #555 we won't be using this approach.

PR #571, opened 2024-09-06
Improve PGP fingerprint handling

koplas:
Warn if no fingerprint is specified and give more details if fingerprint comparison fails. This pull request does not change the API and cannot differentiate if an empty or no fingerprint is provided.


bernhardreiter (approved):
LGTM

JanHoefelmeyer (approved):
LGTM & works

PR #573, opened 2024-09-09
Upgrade to go v1.22

koplas:


bernhardreiter:
Changes look fine, would an upgrade of dependencies be indicated if we update the go version?

bernhardreiter (approved):
LGTM

koplas:
> Changes look fine, would an upgrade of dependencies be indicated if we update the go version?

Some dependencies may need a security update; but this is independent of the go version.

PR #577, opened 2024-09-19
Use .test TLD for integration setup

koplas:
`.local` is reserved for local-area networks, and `.localhost` is reserved for loopback devices. Using `.test` allows easier usage for different test setups.

### DIFF
--- a/docs/development-ca.md
+++ b/docs/development-ca.md
@@ -55,7 +55,7 @@ signing_key
 encryption_key
 non_repudiation
 
-dns_name = "*.local"
+dns_name = "*.test"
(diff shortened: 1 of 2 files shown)


JanHoefelmeyer (approved):
LGTM & works

bernhardreiter:
Additional information for the merging commit:
  * https://www.rfc-editor.org/rfc/rfc2606#section-2 defines the "test." top level domain.
  *  https://www.rfc-editor.org/rfc/rfc6761.html#section-6.2 explains how different implementations can use it

bernhardreiter (approved):
LGTM

PR #584, opened 2024-10-30
improve test setup scripts by adding missing package

bernhardreiter:
Add zip as package to be installed in preparation step as the `make dist` target uses it.


koplas (approved):
LGTM

JanHoefelmeyer (approved):
LGTM  & works

PR #586, opened 2024-11-15
fix: build-in version for release tags 

bernhardreiter:
* Change Makefile to remove the leading `v` from the git tag
   in the case of release tags. Previously this was only done for
   pre-release git tags.


tschmidtb51 (approved):
LGTM

koplas (approved):
LGTM & works

PR #599, opened 2024-11-27
Add Apache 2.0 license to root folder

koplas:
This allows other programs like google/licensecheck to correctly detect the license. This is required to display the documentation in `pkg.go.dev`.

### DIFF
--- a/LICENSE-Apache-2.0.txt
+++ b/LICENSE-Apache-2.0.txt
@@ -1,0 +1,73 @@
+Apache License
+Version 2.0, January 2004
+http://www.apache.org/licenses/
+
+TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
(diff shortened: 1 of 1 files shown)


koplas:
> As this is unneeded by the reuse.standard, we want to improve the software that does the detection pkg.go.dev and prefer it over this cludge.

If we want to provide documentation on pkg.go.dev, we have no other option. See: https://github.com/golang/go/issues/40586#issuecomment-2559897342

bernhardreiter:
> If we want to provide documentation on pkg.go.dev, we have no other option

We can still create the patch in the first place.

koplas:
We cannot control Google's infrastructure.

JanHoefelmeyer (approved):
afaik lgtm

cintek (approved):
LGTM

PR #602, opened 2024-12-04
Warn if no remote validator was specified

### LINKED ISSUE #601: csaf_validator:  Warn if remote_validator is not configured
The validator only validates against the JSON schema if the remote validator is not configured.

In this case a warning should be issued.

koplas:
Prints a warning if no remote validator was configured.

Closes #601

### DIFF
--- a/cmd/csaf_validator/main.go
+++ b/cmd/csaf_validator/main.go
@@ -69,6 +69,8 @@ func run(opts *options, files []string) error {
 				"preparing remote validator failed: %w", err)
 		}
 		defer validator.Close()
+	} else {
+		log.Printf("warn: no remote validator specified")
 	}
 
 	// Select amount level of output for remote validation.
--- a/cmd/csaf_validator/main.go
+++ b/cmd/csaf_validator/main.go
@@ -127,7 +127,6 @@ func run(opts *options, files []string) error {
 		// Check filename against ID
 		if err := util.IDMatchesFilename(eval, doc, filepath.Base(file)); err != nil {
 			log.Printf("%s: %s.\n", file, err)
-			continue
 		}
 
 		// Validate against remote validator.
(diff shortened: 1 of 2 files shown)


s-l-teichmann (dismissed):
LGTM @koplas I have reworked the exit code doc part to make it more clear that the bits are superposed .. Its a bit technical .. so if you have a better wording?

koplas:
> LGTM @koplas I have reworked the exit code doc part to make it more clear that the bits are superposed .. Its a bit technical .. so if you have a better wording?

Most documentations have a table with a Hex Mask, but this is even more technical. This approach seems good enough for me.

bernhardreiter (dismissed):
a return from run() with error condition set with lead to os.exit(1),  but 1 always means `exitCodeSchemaInvalid` so each codepath for this should be checked and either commented or changed.

bernhardreiter:
To get this consistent: 

Do we have a first bit set in all error cases (like the documentation suggests) or only when there is no other error?

Do we have the other bits set in case of general error?

Do we get log output on all errors or only on general errors (as the documentation suggests)?

koplas:
> To get this consistent:
> 
> Do we have a first bit set in all error cases (like the documentation suggests) or only when there is no other error?
> 
> Do we have the other bits set in case of general error?
> 
> Do we get log output on all errors or only on general errors (as the documentation suggests)?

If a general error occurs, all other bits besides the general error bit are unset. I extended the documentation to make this more explicit.

JanHoefelmeyer (commented):
Certain errors will stop the check between local and remote validator without any comment indicating that the remote validation is completely skipped, which may lead to the wrong impression that the remote validation was succesful.

For example, removing the first '{' on a valid document will cause the IDMatchesFilename function to throw an error and the continue will wordlessly skip remote validation. The exit code will then imply that the remote check was succesful.

JanHoefelmeyer (approved):
Other than https://github.com/gocsaf/csaf/pull/602#pullrequestreview-2569593131,  works

cintek (approved):
LGTM

PR #605, opened 2024-12-18
fix: Content-Type header for JSON responses (minor)

bernhardreiter:
* Remove `charset=utf-8` parameter, which is not allowed for JSON, according to rfc8259.

Reported in  https://github.com/gocsaf/csaf/issues/588#issuecomment-2548116924 by @Rafiot


koplas (approved):
LGTM

tschmidtb51 (approved):
LGTM

PR #613, opened 2025-01-29
Errors for lookup checks

JanHoefelmeyer:
Closes https://github.com/gocsaf/csaf/issues/411

Originally, the checker would fail if any of its checks failed. Since only one of the requirements 8-10 is mandatory, whether one of those checks would give out an error or a warning was dependent on the other two checks and they would only fail if all would fail. 

This no longer holds true for the current design of the checker since c7453a644825ddbb1a5c81a8934a82f2af521630. 

The current design of these three checks lead to e.g. https://github.com/gocsaf/csaf/issues/411, so by modernizing the checkWellknownSecurityDNS function as well as its subfunctions this issue is fixed.


koplas (approved):
LGTM

s-l-teichmann (approved):
LGTM

PR #615, opened 2025-02-25
update runner for release to 22.04, other actions and go version, thus also newer glibc

### LINKED ISSUE #614: Upgrade to new runner ubuntu-22.04
As announced by github the `ubuntu-20.04` running used for releases will be phased out.

As we use it from https://github.com/gocsaf/csaf/blob/main/.github/workflows/release.yml
to get a compatible binary with elder distributions, we must also announce that we upgrade the glibc with the next release.

See https://github.com/actions/runner-images/issues/11101
> the image will be fully unsupported by 2025-04-01.

And May 2025 is also the end of standard support for Ubuntu 20.04 LTS, see https://ubuntu.com/about/release-cycle

bernhardreiter:
* Update runner to ubuntu-22.04 which is the eldest to be supported
       by github from 2025-04-01.
 * Update github actions and go version for release and itest workflows.
 * Update current go revision to Readme and development docs.

resolve #614

### DIFF
--- a/.github/workflows/itest.yml
+++ b/.github/workflows/itest.yml
@@ -36,7 +36,7 @@ jobs:
         shell: bash
 
       - name: Upload test results
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
             name: checker-results
             path: |
--- a/.github/workflows/itest.yml
+++ b/.github/workflows/itest.yml
@@ -7,9 +7,9 @@ jobs:
     steps:
 
       - name: Set up Go
-        uses: actions/setup-go@v3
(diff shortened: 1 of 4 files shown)


koplas:
The integration test is failing: https://github.com/gocsaf/csaf/actions/runs/13523335787.

koplas (approved):
LGTM

bernhardreiter:
> LGTM

Does it also work. I suggest you make a dev1 release to check (which can be deleted afterwards)

JanHoefelmeyer (approved):
We should trust our actions here, so LGTM, the works is given through the actions which test the workflow

PR #618, opened 2025-03-03
Fix potential leak of HTTP response body in downloadJSON

marcusperlick:
1. According to standard library the client must close the Body of http response. The function [downloadJSON()](https://github.com/gocsaf/csaf/blob/main/cmd/csaf_aggregator/client.go) in the csaf_aggregator defers the `Body.Close()` after some checks are made and an early return that may return even when `c.Get()` does not return an error. This is where the `res.Body` might not be closed.
2. Besides that either I didn't get the idea behind the return via a call to a closure or it should be simplified.
3. Do not drop the err details from `c.Get()` but give them more context. See Go Blog: _[It is the error implementation’s responsibility to summarize the context.](https://go.dev/blog/error-handling-and-go)_

### DIFF
--- a/cmd/csaf_aggregator/client.go
+++ b/cmd/csaf_aggregator/client.go
@@ -10,6 +10,7 @@ package main
 
 import (
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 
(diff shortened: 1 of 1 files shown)


koplas (approved):
LGTM

s-l-teichmann (approved):
LGTM

PR #621, opened 2025-03-10
Report error in checker if content type is not correct

### LINKED ISSUE #606: Dealing with `charset=utf-8` in application/json responses
Split out from #588:

How should we deal with a `charset=utf-8` parameter to the `content_type: application/json` header in responses? This parameter is not allowed by the current JSON standard (rfc8259).

While rfc9431 recommends to be strict when receiving a non-standard thing for mid term benefits, we also should consider the ease of deployment. 

After talks to @s-l-teichmann and @tschmidtb51, the plan is to 
 * check out how popular webservers like freenginx, Apache and IIS would serve a `.json` file from disk by default.
 * find instructions to easily change this default to something else.

Depending on this we may consider to accept those responses for the downloader and aggregator and be strict on the checker, or be strict with both.

koplas:
Some providers don't correctly return the content type for the advisories. Report an error in the checker to make this defect more visible. The downloader currently displays a warning.

Closes #606

### DIFF
--- a/cmd/csaf_checker/processor.go
+++ b/cmd/csaf_checker/processor.go
@@ -678,9 +678,9 @@ func (p *processor) integrity(
 			continue
 		}
 
-		// Warn if we do not get JSON.
+		// Error if we do not get JSON.
 		if ct := res.Header.Get("Content-Type"); ct != "application/json" {
-			lg(WarnType,
+			lg(ErrorType,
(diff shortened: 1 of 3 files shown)


s-l-teichmann (approved):
LGTM

JanHoefelmeyer (approved):

PR #622, opened 2025-03-10
Extend structured logging usage in aggregator

### LINKED ISSUE #540: Complete transition to structured logging for aggregator
Using structured logging as introduced with #530 is an advantage but removing `--verbose` may be considered an API change and thus would trigger a major release (e.g. 4), maybe changing the logging output is also a change in behavior.

This issues is to consider the implications and plan when to remove `verbose` and how to release the addition. A variant could be to let verbose set a logging level and do a minor release if the log output is not considered part of significant interface.

koplas:
Closes #540

### DIFF
--- a/cmd/csaf_aggregator/config.go
+++ b/cmd/csaf_aggregator/config.go
@@ -264,6 +264,15 @@ func (c *config) privateOpenPGPKey() (*crypto.Key, error) {
 	return c.key, c.keyErr
 }
 
+// httpLog does structured logging in a [util.LoggingClient].
+func httpLog() func(string, string) {
+	return func(method, url string) {
+		slog.Debug("http",
+			"method", method,
+			"url", url)
+	}
+}
+
 func (c *config) httpClient(p *provider) util.Client {
 
 	hClient := http.Client{}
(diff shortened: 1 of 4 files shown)


koplas:
This line may result in unstructured logs and needs to be checked: https://github.com/gocsaf/csaf/blob/3e16741ed5ac4c3037351a20d860f128d6c5d058/cmd/csaf_aggregator/mirror.go#L83

cmd/csaf_aggregator/config.go
```diff
@@ -264,6 +264,15 @@ func (c *config) privateOpenPGPKey() (*crypto.Key, error) {
 	return c.key, c.keyErr
 }
 
+// httpLog does structured logging in a [util.LoggingClient].
+func httpLog() func(string, string) {
```
s-l-teichmann, line 268: No need to have  an inner function here.

s-l-teichmann (dismissed):

JanHoefelmeyer (approved):
LGTM & works

s-l-teichmann (approved):
LGMT

PR #554, opened 2024-07-25
Improve SHA* marking

koplas:
@JanHoefelmeyer Can you check if all requirements of https://github.com/csaf-poc/csaf_distribution/issues/289#issuecomment-1684906779 are met?

Closes #289

PR #623, opened 2025-03-11
fix: fix memory leak on failed http requests in csaf downloader

mgoetzegb:
In many error cases the http response body was not closed as the `defer` statement was only executed after the error handling.

Thus the connection can't be reused. This caused the creation of many goroutines, massively increasing the memory consumption. 

A prominent scenario used to be if a CSAF provider does not provide both sha256 and sha512 checksums. But I think with #554 this is now mostly prevented.

*Note*: I only checked for the csaf downloader, the issue is potentially also present in other components. But I assume there it will have less impact.

Related to issue https://github.com/gocsaf/csaf/issues/591

### DIFF
--- a/cmd/csaf_downloader/downloader.go
+++ b/cmd/csaf_downloader/downloader.go
@@ -484,6 +484,15 @@ nextAdvisory:
 				"error", err)
 			continue
 		}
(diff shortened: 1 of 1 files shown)


koplas:
Superseded by #625.

Issue #479, opened 2023-10-17
set ACAO header to allow cross-origin access for public files


bernhardreiter:
To allow browser based application to access the public files, the HTTP header `Access-Control-Allow Origin: *` must be set.

 * document this requirement in the provider docs
 * add the header to our nginx test setup example.

For more details see https://github.com/csaf-poc/csaf_webview/issues/15
and https://github.com/oasis-tcs/csaf/issues/653 .

PR #624, opened 2025-03-13
feat: add access-control-allow-origin header

bernhardreiter:
.. for better access from web applications.

solve #479

### DIFF
--- a/docs/scripts/setupProviderForITest.sh
+++ b/docs/scripts/setupProviderForITest.sh
@@ -56,11 +56,14 @@ location /cgi-bin/ {
 sudo sed -i "/^server {/a\        include fcgiwrap.conf;" $NGINX_CONFIG_PATH
 
 echo "
-        # For atomic directory switches
-        disable_symlinks off;
+		# For atomic directory switches
+		disable_symlinks off;
 
-        # directory listings
-        autoindex on;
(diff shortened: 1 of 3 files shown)


docs/scripts/setupProviderForITest.sh
```diff
@@ -56,11 +56,14 @@ location /cgi-bin/ {
 sudo sed -i "/^server {/a\        include fcgiwrap.conf;" $NGINX_CONFIG_PATH
 
 echo "
-        # For atomic directory switches
```
koplas, line 59: Is there any reason for switching from spaces to tabs?

koplas (approved):

JanHoefelmeyer (approved):

docs/scripts/setupProviderForITest.sh
```diff
@@ -61,6 +61,9 @@ echo "
 
         # directory listings
         autoindex on;
+
```
bernhardreiter, line 64: This is wrong as the original file uses two tabs, so the result after insertion jumps around. The original version fixes this.

koplas: Please move formatting in a separate commit and use consistent formatting.

PR #625, opened 2025-03-17
Move advisory downloading to download context method

s-l-teichmann:
The response body of a GET request should be closed to avoid memory leaks.
This PR does this.


mgoetzegb:
There is a leftover memory leak, if I see correctly: https://github.com/gocsaf/csaf/blob/5437d8127a8245ea5da2d7162c63c844e16156e9/cmd/csaf_downloader/downloader.go#L845-L853

PR #627, opened 2025-03-17
Use consistent ID naming

koplas:


koplas:
I created and linked an issue. Feel free to provide additional context.

bernhardreiter:
IMO it shouldn't be merged until we do a new major version.

PR #629, opened 2025-03-24
Avoid memory leak

koplas:
Move `resp.Body.Close()` before check of status code.

Reported by @mgoetzegb here: https://github.com/gocsaf/csaf/pull/625#issuecomment-2744067770


JanHoefelmeyer (approved):
LGTM & works

PR #631, opened 2025-03-28
Fix aggregator URL handling

koplas:
Parts of the URL were not path escaped. This results in a wrong URL; if the provider name contains characters that need to be escaped.


JanHoefelmeyer (approved):

s-l-teichmann (approved):
LGTM

PR #633, opened 2025-04-03
Check if canonical url prefix is valid

koplas:
This adds a check to the canonical URL prefix configuration. It detects if an error occurs when the user specifies an invalid URL.

This is a breaking change, as an old configuration that worked with the malformed URL is no longer accepted.

### DIFF
--- a/cmd/csaf_provider/config.go
+++ b/cmd/csaf_provider/config.go
@@ -11,6 +11,7 @@ package main
 import (
 	"fmt"
 	"io"
+	"net/url"
 	"os"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


JanHoefelmeyer (approved):
lgtm

koplas:
One additional note: I could not find a place where some additional input needs to be URL path escaped. But it is also beneficial for future changes to use net/url to concatenate the paths and avoid bugs if further input is introduced.
This can be done after this change, without introducing new breaking changes.

bernhardreiter:
To understand why this is a breaking change:
If there is an old configuration that has an invalid prefix configuration, when would the problem show itself? Or in how far could it work.

I am asking because if this fixes a defect, it can be a patch or minor release, even if it changes the behaviour.

koplas:
> To understand why this is a breaking change: If there is an old configuration that has an invalid prefix configuration, when would the problem show itself? Or in how far could it work.
> 
> I am asking because if this fixes a defect, it can be a patch or minor release, even if it changes the behaviour.

It does not really fix a defect, it only adds validation for the configuration. If an invalid prefix is used it will have a malformed URL in the PMD that can't be accessed. This may result in failure to download advisories from the configured provider.

cintek (approved):
I don't have a fully setup provider but at least I could see that an invalid URL is rejected.

PR #649, opened 2025-06-19
Use JoinPath

koplas:
This avoids issues where parts of the URL are discarded.

This currently fails as we don't track the correct base URL.


JanHoefelmeyer (approved):
LGTM & works

cintek (approved):
LGTM

PR #651, opened 2025-06-20
improve calculated version numbers

bernhardreiter:
for modified git workspaces a `-modified` is added to the semantic version
  in the makefile.


koplas:
Using `--broken` seems to break #650.

koplas:
Run `mv .git .git-tmp` and then `make`. The result is the same when you download the auto generated tarball.

koplas (approved):

JanHoefelmeyer (approved):

PR #680, opened 2025-09-12
Rename workflow go_legacy to "Go Test (oldstable)"

bernhardreiter:
so it is distinct from the other "Go" workflow


cintek (approved):
LGTM

PR #689, opened 2025-10-08
improve  `LoadCertificate` unit test

mgoetzegb:
## What

_~fix~ improve_ `LoadCertificate` unit test

replaced certificate with invalid dns name, which is rejected by the  stdlib of Go version ==1.25.2.

Change in Go introduced by https://github.com/golang/go/issues/75715

_Update -10-17:_ go 1.25.3 did not have the problematic change, see  https://github.com/cockroachdb/go/commit/961e9c3dd4c58115ca7466104670e5ddc79150e2 (and testing below)).

## Why

Unit test will be failing as soon go `1.25.2` will be used by the CI otherwise.

### DIFF
--- a/internal/certs/data/generate_cert.sh
+++ b/internal/certs/data/generate_cert.sh
@@ -1,4 +1,12 @@
 #!/bin/sh
+#
+# This file is Free Software under the Apache-2.0 License
+# without warranty, see README.md and LICENSES/Apache-2.0.txt for details.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# SPDX-FileCopyrightText: 2025 German Federal Office for Information Security (BSI) <https://www.bsi.bund.de>
+# Software-Engineering: 2025 Intevation GmbH <https://intevation.de>
 
 # Script to generate the certificates used for testing.
 


mgoetzegb:
I took a look at the existing scripts, however none of them generate a password protected key (not protected keys work fine).

I attached the full (non-working) script using `certtool` : [generate_cert.sh](https://github.com/user-attachments/files/22923631/generate_cert.sh)

The tests fail with 
```sh
go test ./internal/certs/...
--- FAIL: TestLoadCertificates (0.00s)
    certs_test.go:49: Failure: Couldn't load supposedly valid encrypted certificate. Got error: x509: no DEK-Info header in block
FAIL
FAIL    github.com/gocsaf/csaf/v3/internal/certs        0.003s
FAIL
```
This is quite certainly due to the used golang function not being able to interpret the used format `PKCS8` (see stackoverflow: https://stackoverflow.com/a/32994842). 

And unlike with `openssl`  `--traditional` flag I did not find a way to generate an encrypted client key with `certtool` in a compatible format.

There is already a [disclaimer in the csaf downloader readme](https://github.com/gocsaf/csaf/blob/05eae0a9ae6f945f946dd815ec7463da5ba0a7de/docs/csaf_downloader.md#beware-of-client-cert-passphrase) regarding the handling of password protected client certificates. So that might be a fine behavior and not an issue.

So feel free to build on the existing script, if you find a way to use `certtool` or if the implementation is adjusted.

bernhardreiter:
@mgoetzegb thanks again, good work on checking the situation!

PR #692, opened 2025-10-17
Improve LoadCertificate unit test

bernhardreiter:
This supercedes #689, see there for a detailed discussion.

 * exchange certificates with new one that have better dns names embedded.
 * add a script to document how the certificates were created.
 * improve message in case of test failures

Thanks to @mgoetzegb  who did the important work on #689.


bernhardreiter:
Testing: Should be enough to run the go tests in `internal`.

go 1.25.2. is the one that has the stricter test for the domain names in the certificate. 1.25.3 relaxed that test again,

s-l-teichmann (approved):
LGTM

PR #696, opened 2025-10-23
silence revive linter warnings

### LINKED ISSUE #682: silence linter ID warning
So that we do not need to change the API

Add new issue for potential API change.

bernhardreiter:
..   that we cannot or do not want to fix yet

For  better naming for variables like `ProductIds`, we must wait for the next release.
With `util` we may just keep it.

This is based on #695, so merge #695 first.

close #682


JanHoefelmeyer (approved):
Doesn't silence all warnings, but many without introducing new ones so LGTM & works

Issue #669, opened 2025-07-31
csaf_uploader: Exit code is 0 despite error


wagner-intevation:
With the error described in #666 the `csaf_uploader` exits with code 0 although the request did not succeed:

```
$ ./csaf-3.3.0-gnulinux-amd64/bin-linux-amd64/csaf_uploader --action create --url http://127.0.0.1/cgi-bin/csaf_provider.go --password password
2025/07/31 11:08:02 WARN No config file found. Maybe you want to specify one or store it in a respective default location locations="[~/.config/csaf/uploader.toml ~/.csaf_uploader.toml csaf_uploader.toml]"
2025/07/31 11:08:02 Create failed: 400 Bad Request
Errors:
	mkdir output: permission denied
2025/07/31 11:08:02 No CSAF files given.
$ echo $?
0
```

In case of no success, the exit code must be non-zero. Otherwise, using the `csaf_uploader` tool in scripts is a pain in the ass, and curl is the more viable option.

PR #698, opened 2025-11-05
Fixes missing TLPLabel type for TLPLabel constants

timbastin:
Just fixes the bug, that TLPLabels were untyped.

### DIFF
--- a/csaf/models.go
+++ b/csaf/models.go
@@ -26,23 +26,23 @@ type TLPLabel string
 
 const (
 	// TLPLabelUnlabeled is the 'UNLABELED' policy.
-	TLPLabelUnlabeled = "UNLABELED"
+	TLPLabelUnlabeled TLPLabel = "UNLABELED"
 	// TLPLabelWhite is the 'WHITE' policy.
-	TLPLabelWhite = "WHITE"
+	TLPLabelWhite TLPLabel = "WHITE"
 	// TLPLabelGreen is the 'GREEN' policy.
-	TLPLabelGreen = "GREEN"
+	TLPLabelGreen TLPLabel = "GREEN"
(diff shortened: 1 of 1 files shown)


timbastin:
> Thank you for your contribution, but we cannot do this change at the moment as it will impose an API change. There might exists some external projects which rely on these constants being untyped.
> 
> Furthermore the PR is incomplete. As you can see the build is failing.
> 
> We are going to plan some API changes for v4 which for example will include support for CSAF 2.1. We will consider to make the constants typed in this phase..

Alright - so should I fix the build or not? Since you mentioned it and closed the pr in the same minute 🙂

bernhardreiter:
Hi @timbastin 

> Alright - so should I fix the build or not?

no -  it is not necessary - thanks for the offer! :)

Background is: We will consider the proposed change in principle at a much later point in time, I've created #669 for it, but we cannot yet say yet, if it is better to keep the situation as as it is or change it. ("Untyped" consts do have advantages in Go as well...)

bernhardreiter:
> Personally, I prefer untyped constants

Out of curiosity, what was the rationale that you have considered the initial situation a "bug" then? :)

timbastin:
I initially considered the situation a “bug” because almost all other constants in the project were typed. At the time, it wasn’t immediately clear to me, especially when using the project as a library, that this particular constant should be an exception. My priority was to ensure a consistent API for users, which I value more highly than the use of untyped constants.

bernhardreiter:
>  My priority was to ensure a consistent API for users, which I value more highly than the use of untyped constants.

Thanks for the additional explanation. Yes we also value consistency so this is why we will consider making the change in #669 for the next major release.

PR #710, opened 2025-12-17
fix: engine is invalid when name is missing

### LINKED ISSUE #709: Engine validation fails when version is empty
Hi, I am facing an issue and I am not sure if it is me not using the library correctly or if this is an issue.

I am parsing a document that contains the following value for `document.tracking.generator`:

```json
{
  "document": {
    ... 
    "tracking": {
      "current_release_date": "2025-12-04T10:14:56Z",
      "generator": {
        "date": "2025-12-04T10:14:56Z",
        "engine": {
          "name": "pyCSAF"
        }
      },
      ...
  },
```

The Go struct for [engine](https://github.com/gocsaf/csaf/blob/main/csaf/advisory.go#L267) has 
```
Version *string `json:"version,omitempty"`
```
which seems matching [the spec](https://docs.oasis-open.org/csaf/csaf/v2.0/csaf_json_schema.json) which states that only the name is required 👇 

```json
"engine": {
	"title": "Engine of document generation",
    "description": "Contains information about the engine that generated the CSAF document.",
    "type": "object",
    "required": [
    	"name"
    ],
    ...
} 
```

But the `Validate` [function](https://github.com/gocsaf/csaf/blob/main/csaf/advisory.go#L894) returns an error when the version is `nil`.

When running the validate command, the csaf file appears to be valid, so I was leanning toward a bug. The `Validate` should not return an error when the version is nil.

benja-M-1:
Hello 👋 , this PR fixes https://github.com/gocsaf/csaf/issues/709

According to the [specification](https://docs.oasis-open.org/csaf/csaf/v2.0/csaf_json_schema.json), the validation of the `tracking.generator.engine` should check the name is not nil rather than the version which is optional. This is what the PR changes.

### DIFF
--- a/csaf/advisory.go
+++ b/csaf/advisory.go
@@ -891,8 +891,8 @@ func (rs Revisions) Validate() error {
 
 // Validate validates an Engine.
 func (e *Engine) Validate() error {
-	if e.Version == nil {
-		return errors.New("'version' is missing")
+	if e.Name == nil {
+		return errors.New("'name' is missing")
 	}
 	return nil
 }
--- a/csaf/advisory_test.go
+++ b/csaf/advisory_test.go
@@ -14,11 +14,12 @@ func TestLoadAdvisory(t *testing.T) {
 		name    string
 		args    args
 		wantErr bool
-	}{{
-		name:    "Valid documents",
-		args:    args{jsonDir: "csaf-documents/valid"},
-		wantErr: false,
-	},
+	}{
+		{
+			name:    "Valid documents",
+			args:    args{jsonDir: "csaf-documents/valid"},
+			wantErr: false,
+		},
 		{
 			name:    "Garbage trailing data",
 			args:    args{jsonDir: "csaf-documents/trailing-garbage-data"},
--- a/testdata/csaf-documents/valid/advisory-tracking-generator-no-version.json
+++ b/testdata/csaf-documents/valid/advisory-tracking-generator-no-version.json
@@ -1,0 +1,169 @@
+{
+  "document": {
+    "category": "csaf_vex",
+    "csaf_version": "2.0",
+    "distribution": {
+      "tlp": {
+        "label": "WHITE",
+        "url": "https://www.first.org/tlp/v1/"
+      }
+    },
+    "notes": [
+      {
+        "category": "summary",
+        "title": "Test document summary",
+        "text": "Auto generated test CSAF document"
+      }
+    ],
+    "publisher": {
+      "category": "vendor",
+      "name": "ACME Inc.",
+      "namespace": "https://www.example.com"
+    },
+    "title": "Test CSAF document",
+    "tracking": {
+      "current_release_date": "2020-01-01T00:00:00Z",
+      "generator": {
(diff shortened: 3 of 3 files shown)


csaf/advisory.go
```diff
@@ -891,8 +891,8 @@ func (rs Revisions) Validate() error {
 
 // Validate validates an Engine.
 func (e *Engine) Validate() error {
-	if e.Version == nil {
-		return errors.New("'version' is missing")
+	if e.Name == nil {
```
bernhardreiter, line 894: I wonder if this is necessary, as the schema validation we do, should catch a missing name. So maybe the function can be removed altogether?

benja-M-1: That works for me! Doing it

benja-M-1: However, when creating an Advisory "manually" and calling the Validate function on it, there are no checks (that's how I spotted the problem), same when calling the `LoadAdvisory` function.

benja-M-1: And I am thinking, that's a backward compatibility break. The function is exposed, which means if someone uses it then their code will break when updating the next version of the lib. 
Are you still sure you want to remove if? :)

bernhardreiter: > sure you want to remove if?

No, just thinking aloud.

>  that's a backward compatibility break.

We could empty the function, so the API stays the same.  

Removing the wrong check is a fix, so we could do a patch release.

bernhardreiter: > there are no checks

Does the schema validation catch the missing name in that situations? That would be "a check", wouldn't it?

benja-M-1: I tested with the main branch, not the branch with the fix.

>   * /document/tracking/generator/engine: missing property 'name'

That's the output when the name is missing. 

>   * /document/tracking/generator/engine/name: minLength: got 0, want 1
 
And that's the output when the name is empty.

When the version is missing it does not fail as it does when loading a file.

bernhardreiter: > When the version is missing it does not fail as it does when loading a file.

What do you mean in particular? It would be okay, if it does not fail, I think.

benja-M-1: Sorry I wasn't very clear. What I meant is:
- `go run ./cmd/csaf_validator/main.go` does not fail, which is 👍 
- `csaf.LoadAdvisory(path)` does fail and the changes I propose fixe it

bernhardreiter:
> go run ./cmd/csaf_validator/main.go does not fail, which is 👍

But doesn't this mean that it does not call Validate() and probably should? *confused*
Guess I'll have a look at some tests or @s-l-teichmann tomorrow.

benja-M-1:
> But doesn't this mean that it does not call Validate() and probably should? confused

No, the validator commad uses the jsonschema to validate. They don't use the same code.

bernhardreiter (approved):
LGTM

s-l-teichmann (approved):
LGTM & works

s-l-teichmann (approved):
LGTM

PR #731, opened 2026-06-03
Stop checking if there is no valid PMD

### LINKED ISSUE #726: If no PMD is found: stop checking
In case no PMD is found, the domain cannot be a CSAF provider.
So processing should stop.

Otherwise the report has "PASS" results that are missleading, like when trying to scan `example.com`:

Image

They are missleading, because its possible to have an empty provider with everything in place, but no files.
Stopping the processing after the ERROR that no PMD was found for the domain shall prevent missleading PASSes.

s-l-teichmann:
Generate a report for these domains.

Resolves #726

### DIFF
--- a/cmd/csaf_checker/processor.go
+++ b/cmd/csaf_checker/processor.go
@@ -257,11 +257,9 @@ func (p *processor) run(domains []string) (*Report, error) {
 			p.badProviderMetadata.use()
 			p.badProviderMetadata.error("Could not parse the Provider-Metadata.json of: %s", d)
 			// If we don't have a valid PMD there is no hope to do any further checking.
-			req := &Requirement{
-				Description: "Invalid provider metadata",
-			}
-			req.Append(p.badProviderMetadata)
-			domain.Requirements = append(domain.Requirements, req)
+			(&providerMetadataReport{
+				baseReporter{description: "Invalid provider metadata"},
+			}).report(p, domain)
 			report.Domains = append(report.Domains, domain)
 			continue
 		}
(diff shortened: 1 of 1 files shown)


cmd/csaf_checker/processor.go
```diff
@@ -251,17 +251,23 @@ func (p *processor) run(domains []string) (*Report, error) {
...
 		p.reset()
 
+		domain := &Domain{Name: d}
 		if !p.checkProviderMetadata(d) {
 			// We need to fail the domain if the PMD cannot be parsed.
 			p.badProviderMetadata.use()
 			p.badProviderMetadata.error("Could not parse the Provider-Metadata.json of: %s", d)
-
+			// If we don't have a valid PMD there is no hope to do any further checking.
+			(&providerMetadataReport{
```
4echow, line 260: When trying to understand this pattern to call the providerMetadataReport `report` method I noticed that it's the only reporter type ending on `Report` not `Reporter`, since it breaks the pattern of the other reporters I assume it's just a typo?

4echow (approved):
mentioned the possible typo, otherwise PR lgtm

bernhardreiter:
> otherwise PR lgtm

Am I right that you haven't run it?

4echow:
> > otherwise PR lgtm
> 
> Am I right that you haven't run it?

Only with a non-provider (10.100.0.1), here a re-run, the description doesn't appear anywhere else in the checker
and @s-l-teichmann used the already existing `providerMetadataReport` method to report, so I assume it already worked beforehand.

```bash
./csaf_checker 10.100.0.1
2026/06/09 22:36:45 WARN No config file found. Maybe you want to specify one or store it in a respective default location locations="[~/.config/csaf/checker.toml ~/.csaf_checker.toml csaf_checker.toml]"
{
  "domains": [
    {
      "name": "10.100.0.1",
      "requirements": [
        {
          "num": 0,
          "description": "Invalid provider metadata",
          "messages": [
```

JanHoefelmeyer (approved):
LGTM & works

Issue #734, opened 2026-06-04
CSAF uploader should support GPG-Agent


tommylehmann:
# Problem

CSAF uploader has to options to handle Signatures of CSAF Documents. Eigher expect them to be externally signed, which requires an additional step or use an OpenPGP private key file. Using an OpenPGP private key file requires one to duplicate the private key storage into OpenGPGs keyring and an exported version for csaf_uploader. Also the Signing key can't be stored on a secure hardware key storage like a yubikey or smartcard.

# Solution

Csaf_uploader should have an option to call the gpg-agent to do the signing. That would simplify private key distribution if the key is also used for other tasks like email and would benefit from everything gpg does to secure the private key.

bernhardreiter:
Hi @tommylehmann,
thanks for the suggestion!

Technically csaf_uploader does not use GnuPG but https://github.com/ProtonMail/go-crypto.
And go-crypto does not support GnuPG's agent protocol.

csaf_uploader can be used in a script which could bundle both steps, signature by GnuPG and then upload. Would that be a solution for you?

tommylehmann:
@bernhardreiter that would be an option, but I think key handling should be in the csaf_uploader. Otherwise it would make sense to also remove the existing gpg signing option and expect everything externally signed. I just made a pull request.

bernhardreiter:
Externally signing the CSAF files is the recommendation. In some situations and for testing it is not ideal - so we have more options.

bernhardreiter:
Yesterday we've talked within the current gocsaf devs about this issue and #735 .

@tommylehmann can you tell us more about your use case: What tools are you using to create your CSAF documents?
How many do you create and upload in one go? On what operating systems/platform are you using these tools?

In #735 you wrote:
> For me it was - until now : Creating a CSAF Document with Secvisogram, exporting the JSON, pushing the JSON to trusted provider using uploader.

So far we think that adding an optional GnuPG dependency and code to do external signing is not worth the extra code and its maintaining efforts . And if we would add , we would design it more flexible with a hook and think about cross platform possibilities.

Depending on your use case and platform, we think that more documentation and example scripts could help you to smooth your process. We can try to provide these.

bernhardreiter:
Adding the GnuPG signing to Secvisogram is not a good solution either, I agree with that. ;-)

> I could write a wrapper around csaf_uploader but then I end up having just more complexity.

If we provide an example wrapper for GNU/Linux systems in bash or python, would that help you?
[...]

PR #735, opened 2026-06-04
#734 feat(csaf_uploader): add optional signing via system gpg/gpg-agent

tommylehmann:
Implements #734

Add a --gpg mode that produces the armored detached signature by shelling out to the system gpg binary, delegating passphrase/PIN handling to gpg-agent/pinentry. This lets the signing key live on a smartcard, YubiKey or other hardware token instead of being an exported key file loaded into the uploader process.

New flags:
  --gpg          enable signing via the system gpg
  --gpg-user     signing key identity for gpg --local-user (optional)
  --gpg-binary   path to the gpg executable (default: gpg)

Signing is routed through a small internal signer interface (keyRingSigner for the existing in-process gopenpgp path, gpgSigner for the new exec path), so uploadRequest has a single call site per mode. --gpg is mutually exclusive with -k/--key and -x/--external_signed; -P/-I are ignored in gpg mode. The on-the-wire signature field is unchanged (ASCII-armored detached OpenPGP signature), so csaf_provider verifies it without changes.

The existing no-sign, file-key and external-signed modes are unchanged.

Includes unit tests (gpg invocation driven by a stub binary) and a gated integration test that created a temp GPGHOME, generates a fresh gpg key, signs with a real gpg and verifies using the gopenpgp library.

### DIFF
--- a/cmd/csaf_uploader/config.go
+++ b/cmd/csaf_uploader/config.go
@@ -23,9 +23,10 @@ import (
 )
 
 const (
-	defaultURL    = "https://localhost/cgi-bin/csaf_provider.go"
-	defaultAction = "upload"
-	defaultTLP    = "csaf"
+	defaultURL       = "https://localhost/cgi-bin/csaf_provider.go"
+	defaultAction    = "upload"
+	defaultTLP       = "csaf"
+	defaultGPGBinary = "gpg"
 )
 
 // The supported flag config of the uploader command line
@@ -37,6 +38,10 @@ type config struct {
 	TLP            string `short:"t" long:"tlp" choice:"csaf" choice:"white" choice:"green" choice:"amber" choice:"red" description:"TLP of the feed" toml:"tlp"`
 	ExternalSigned bool   `short:"x" long:"external_signed" description:"CSAF files are signed externally. Assumes .asc files beside CSAF files." toml:"external_signed"`
 	NoSchemaCheck  bool   `short:"s" long:"no_schema_check" description:"Do not check files against CSAF JSON schema locally." toml:"no_schema_check"`
+
+	GPG       bool    `long:"gpg" description:"Sign CSAF files via the system gpg binary (delegates passphrase/PIN to gpg-agent)" toml:"gpg"`
+	GPGUser   *string `long:"gpg-user" description:"Signing key identity for gpg --local-user (fingerprint/email/keyid)" value-name:"IDENTITY" toml:"gpg_user"`
+	GPGBinary string  `long:"gpg-binary" description:"Path to the gpg executable" value-name:"GPG-BIN" toml:"gpg_binary"`
 
 	Key              *string `short:"k" long:"key" description:"OpenPGP key to sign the CSAF files" value-name:"KEY-FILE" toml:"key"`
 	Password         *string `short:"p" long:"password" description:"Authentication password for accessing the CSAF provider" value-name:"PASSWORD" toml:"password"`
@@ -55,7 +60,7 @@ type config struct {
 
 	clientCerts []tls.Certificate
 	cachedAuth  string
-	keyRing     *crypto.KeyRing
+	signer      signer
 }
 
 // iniPaths are the potential file locations of the the config file.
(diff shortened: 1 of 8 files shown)


bernhardreiter:
Hi, thanks for the contribution!

It will take us a while to review it.

As this is a bit of code since yesterday, did you apply LLM help?
(I wonder because the copyright on new files is BSI, was that intentional?)

Usually programmatic access to GnuPG should be done via GPGME, for a more robust messaging between the calling application and the crypto backend. I haven't checked out the available options like https://github.com/proglottis/gpgme so far.

Best Regards,
Bernhard

tommylehmann:
Yes I did use LLMs. But honestly I would have copied the header non the less. There is no contribution.md or something that tells me what to use.

To GPGME. Apparently I was not aware of GPGME was for programmatic access. It would introduce an additional dependency. Calling gpg from commandline does not do this and it just fails if gpg is not available. I will have look into gpgme.

tommylehmann:
# Summary

Using GPGME instead of GPG Shell Callout would increase build complexity and introduce a runtime dependency for an optional feature. Using a GPG via Shall Callout is not that obscure as everyone using Git with Signed Commit and Tags already got in touch with it.

# Details

I did some research regarding GPGME. It seams like [roglottis/gpgme](https://github.com/proglottis/gpgme) is the only Go implementation. But its full implementation its a wrapper around the GPGME C Library. I would not recommend using it for the following reasons.

1. It introduces an additional build time dependency. [roglottis/gpgme](https://github.com/proglottis/gpgme) compiles with CGO against the GPGME C Library which requires a C compiler and GPGME Headerfiles. The CSAF Tooling is currently pure Go which makes cross compiling cheap. The C dependency would require a C Cross Compile Chain for each platform rising the complexity for the build.
2. It would introduce an additional runtime dependency. Since [roglottis/gpgme](https://github.com/proglottis/gpgme)  just dynamically links its wrapper against libgpgme the executable will fail during startup at load time if the system where its executed does not have libgpgme installed. This could break current deployments.
3. Other Projects like [git](https://github.com/git/git/blob/9ac3f193c05c2237e2b14ebaa1149e9fc8a1abe0/gpg-interface.c#L96) do it in the same way. I could not find any justification for this but its probably for the same reason like 2. That's also the reason I have not thought about a different approach initially.

bernhardreiter:
> Yes I did use LLMs.

Thanks for indicating it!
Obviously the software-engineering line is not correct.

We will have to talk about this in our team, it is the first LLM generated contribution we get for this repo. In other contexts LLM based contributions have quality problems, sometimes subtle, so review takes more time and efforts.

 > It would introduce an additional dependency.

Compared to the robustness, GPGME is worth it, as it is part of GnuPG, it is not really a new dependency, the big dependency is GnuPG.  (The reasoning above also seems LLM generated, too, and is beside the point, or flawed to say so. Please avoid sending us text that has theses flaws.)

tommylehmann:
Der obige Text ist tatsächlich nicht LLM generiert, sondern komplett per Hand geschrieben. Ich kann die Skepsis allerdings verstehen.

bernhardreiter:
Thanks for looking into a GPGME based implementation! (And sorry for the suspicion, it is getting harder these days to see which text comes from a human that is thinking and searching for an answer or from a probabilistic process. :/ )

You maybe correct that introducing libgpgme as C library may add a hard requirement, I haven't fully checked yet whether it can be made optional in an easy way from Go. 

And the bigger question is: Isn't this a lot of code and complexity for this feature? A feature that potentially is better placed on the other side of the process.

tommylehmann:
What do you mean with the other side of the process?
For me it was - until now : Creating a CSAF Document with Secvisogram, exporting the JSON, pushing the JSON to trusted provider using uploader.
So the Other side in my understanding would be the trusted provider which also as the ability to sign the JSON. From my point of view letting the trusted provider sign the JSON is a far higher risk because a somehow public facing server than has a secret key.
On the different other side is Secvisogram and i don't think adding a GPG to a Webapp is in any way better than keeping it fully local. I could write a wrapper around csaf_uploader but then I end up having just more complexity. I really considdered this way since the GPG handling bothers me since 2 years.

bernhardreiter:
For the reasons outlined in #734 we suggest to not pursue this code addition in the current form.
Let us continue the more general discussion in #734.

s-l-teichmann:
We are currently evaluating
https://github.com/gocsaf/csaf/pull/738
which adds an option to call an external signing tool.

This is a more general approach that keeps the details of the signing out of the uploader.
With this you can write a wrapper script around `gpg` which can use its agent to do the actual signing.
We are going to provide an example how to use this.

tommylehmann:
Ok, I will close this Pull Request in Favor of #738 .

PR #736, opened 2026-06-10
Add utf8 validation for files in checker, downloader and validator

### LINKED ISSUE #716: Flag badly encoded documents
Do we detect badly encoded documents and treat them accordingly?

https://github.com/ISDuBA/ISDuBA/issues/1470 has the example document [opensuse-su-2025_14737-1.json](https://github.com/user-attachments/files/27204610/opensuse-su-2025_14737-1.json) which contains an invalid characters, so it is not UTF-8 encoded.

Go's json parser changes those characters within strings to the unicode replacement character, but does not raise an error.

* [ ] check if our downloader, checker and validator will accept those documents without warning or error.
* [ ] decide what would be a good behaviour? E.g. download acceptance by default, checker reject by default?
* [ ] estimate and then potentially implement an improvement.

4echow:
partially resolves #716

### DIFF
--- a/cmd/csaf_checker/processor.go
+++ b/cmd/csaf_checker/processor.go
@@ -27,6 +27,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"github.com/gocsaf/csaf/v3/internal/misc"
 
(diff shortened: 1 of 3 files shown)


JanHoefelmeyer (approved):
LGTM & works

s-l-teichmann (approved):
LGTM

PR #738, opened 2026-06-15
Uploader: add option to call external signing tool

s-l-teichmann:
This PR adds an option to call an external signing tool:

```
-X, --signing_tool=                    Tool to sign a file externally
      --signing_tool_timeout=   Timeout for the external signing tool
```

The idea is to enable using an external tool like `gpg` to sign
an advisory before uploading it to the provider.
The advisory is piped to the tool via stdin. The signature is captured from stdout.

TODO:
- [x] Documentation
- [x] Example
- [x] Test

### DIFF
--- a/cmd/csaf_uploader/config.go
+++ b/cmd/csaf_uploader/config.go
@@ -13,6 +13,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"time"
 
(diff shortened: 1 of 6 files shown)


4echow (approved):
lgtm + works

JanHoefelmeyer (approved):
LGTM

PR #742, opened 2026-06-30
Add log for redundant PMDs in security.txt

### LINKED ISSUE #719: Reporting found, but unused PMDs for CSAF 2.0 security.txt
The security.txt can have several PMDs.

csaf_downloader, checker and ISDuBA only use one (the first valid one as far as I know).

This is an issue to report about other PMDs, that were found, but are not used.
So users of the command line tool or online tool can see that there is an unused PMD and also see the warning that it should be removed as soon as possible.

(This is the less common subcase of #712)

### reference

https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#718-requirement-8-securitytxt

> It is possible to advertise more than one provider-metadata.json by adding multiple CSAF fields, e.g. in case of changes to the organizational structure through merges or acquisitions. However, this SHOULD NOT be done and removed as soon as possible. If one of the URLs fulfills requirement 9, this MUST be used as the first CSAF entry in the security.txt.

4echow:
partially resolves #719 in csaf_checker

### DIFF
--- a/cmd/csaf_checker/processor.go
+++ b/cmd/csaf_checker/processor.go
@@ -1363,10 +1363,15 @@ func (p *processor) checkSecurityFolder(ctx context.Context, folder string) stri
 
 	u, err := func() (string, error) {
 		defer res.Body.Close()
-		lines, err := csaf.ExtractProviderURL(res.Body, false)
+		lines, err := csaf.ExtractProviderURL(res.Body, true)
 		var u string
 		if len(lines) > 0 {
 			u = lines[0]
+			if len(lines) > 1 {
+				for _, line := range lines[1:] {
+					log.Printf("WARN: Unused PMD in security.txt: %s\n", line)
+				}
+			}
 		}
 		return u, err
 	}()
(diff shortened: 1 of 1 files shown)


s-l-teichmann (dismissed):
LGTM

cmd/csaf_checker/processor.go
```diff
@@ -1529,10 +1529,13 @@ func (p *processor) checkSecurityFolder(ctx context.Context, folder string) stri
 
 	u, err := func() (string, error) {
 		defer res.Body.Close()
-		lines, err := csaf.ExtractProviderURL(res.Body, false)
+		lines, err := csaf.ExtractProviderURL(res.Body, true)
 		var u string
 		if len(lines) > 0 {
 			u = lines[0]
+			for _, unused := range lines[1:] {
+				log.Printf("WARN: Unused PMD in security.txt: %s\n", unused)
```
JanHoefelmeyer, line 1549: Shouldn't this be a warning added to the security.txt-reporter?

s-l-teichmann: Valid point. [d1cf49d](https://github.com/gocsaf/csaf/pull/742/commits/d1cf49d4a631e53b8d2b4be3e89f674ebe6003c9) adds it to the respective reporter.

I think the reason was to report this early. Reporters are only applied to the result document.
So I left in the log logging for now.

s-l-teichmann (approved):
LGTM

JanHoefelmeyer (approved):
LGTM & works

PR #744, opened 2026-07-15
csaf: drop null entries when loading a ROLIE feed

arpitjain099:
A `null` inside a ROLIE feed's `entry` array parses fine (`StrictJSONParse` is happy with it) and lands in `FeedData.Entry` as a nil `*Entry`. Nothing downstream expects that, so the first consumer to touch it panics.

```
{"feed":{"id":"x","title":"t","updated":"2020-01-01T00:00:00Z","entry":[null]}}
```

Against that feed, `Entries` hands the nil straight to the callback and the downloader's callback in `csaf/advisories.go` does `time.Time(entry.Updated)` right away:

```
panic: runtime error: invalid memory address or nil pointer dereference
	github.com/gocsaf/csaf/v3/csaf.(*ROLIEFeed).Entries(...)
		csaf/rolie.go:237
```

`SortEntriesByUpdated` and `EntryByID` go the same way. These feeds come off remote providers, so csaf_downloader/csaf_aggregator/csaf_checker all reach it with input they don't control. The aggregator case is the annoying one since one bad provider takes down the whole run.

I dropped the nil entries in `LoadROLIEFeed` instead of guarding each consumer, that way the invariant holds for external callers too. Test covers the null-only feed and a null mixed in with a real entry (`sort.Slice` skips the comparator on a length-1 slice, so the sort path needs two).

`go test ./csaf/...` passes.

### DIFF
--- a/csaf/rolie.go
+++ b/csaf/rolie.go
@@ -11,6 +11,7 @@ package csaf
 import (
 	"encoding/json"
 	"io"
+	"slices"
 	"sort"
 	"time"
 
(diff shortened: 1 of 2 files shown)


bernhardreiter:
Hi @arpitjain099, thanks for your submission, we will take a look. :)

s-l-teichmann (approved):
LGTM

4echow (approved):
lgtm

PR #745, opened 2026-07-16
Prevent large buffers used in dowloading

s-l-teichmann:
The checker, dowloader and aggregator are fetching advisories from the net.
To reduce memory strain the memory allocated for one file is hold and re-used
for the next advisories.

The implementation lead to  using more memory than needed. Once a large advisory is
loaded the corresponding large chunk of memory is hold till the end of the program.
Having more downloading workers the problem multiples.

This PR integrates a buffer recycling between downloading workers that only
helds memory blocks up to 4 MiB. Larger chunks are not reused. This is
sensible as larger advisories are rare compared to the majority of small ones.

### DIFF
--- a/cmd/csaf_aggregator/full.go
+++ b/cmd/csaf_aggregator/full.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/gocsaf/csaf/v3/csaf"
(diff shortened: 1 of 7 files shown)


JanHoefelmeyer (approved):
LGTM & works

4echow (approved):
lgtm

